PR #5, opened 2025-05-14
Create a template for every OVA VM

art-dsit:
This greatly improves the performance when creating VMs at the start of an eval, if the eval has already been run at least once.


README.md
```diff
@@ -185,6 +185,14 @@ from a Proxmox server in qcow2 format (instructions for this can be found online
 Once you have the disk images locally, you can use the convenience script `src/proxmoxsandbox/scripts/ova/convert_ova.sh`
 to convert it into an OVA.
 
+This provider creates a template VM for every OVA-type VM specified in an eval.
+Next time you run the eval, a linked clone of the template VM will be created.
+This is for performance. If you change the OVA, as long as the filesize changes, 
+a new template VM will be created.
+
+These template VMs are *not* cleaned up because that needs to happen outside
```
Jay-Bailey-DSIT, line 193: How important is it to clean them up after an eval is done? If we just leave them around, what happens?

src/proxmoxsandbox/_impl/qemu_commands.py
```diff
@@ -229,67 +230,107 @@ async def create_from_backup() -> None:
                     f"Not supported: {type(vm_config.vm_source_config.ova)}"
                 )
             if isinstance(vm_config.vm_source_config.ova, Path):
-                await self.storage_commands.upload_file_to_storage(
-                    file=vm_config.vm_source_config.ova,
-                    content_type="import",
-                )
+                ova_size = vm_config.vm_source_config.ova.stat().st_size
```
willpayne23, line 233: Why not use some kind of hash? I can imagine someone eventually hitting the different file same size collision and ripping their hair out (even with your line in the README)

README.md
art-dsit: They take up disk space and clog up the list of VMs in the UI

src/proxmoxsandbox/_impl/qemu_commands.py
art-dsit: I was considering this but didn't want to pay the cost of hashing the ova file every time - it can add up to tens of seconds for md5sum for a multi-GB file. Perhaps file size + last modified timestamp?

I'll also investigate faster, non-cryptographic, hashes.

art-dsit: I've just realised there is another problem, which is that the [OVA upload itself ](https://github.com/UKGovernmentBEIS/inspect_proxmox_sandbox/blob/main/src/proxmoxsandbox/_impl/storage_commands.py#L42) won't reupload an OVA with an identical name - it doesn't even check the filesize! This needs a more robust, general, solution for both places that the OVA content is implicitly cached.

art-dsit: > I've just realised there is another problem, which is that the [OVA upload itself ](https://github.com/UKGovernmentBEIS/inspect_proxmox_sandbox/blob/main/src/proxmoxsandbox/_impl/storage_commands.py#L42) won't reupload an OVA with an identical name - it doesn't even check the filesize! This needs a more robust, general, solution for both places that the OVA content is implicitly cached.

This is fixed now in this PR

PR #6, opened 2025-05-16
Attempt to increase file upload chunk size for Proxmox 8.4 (not working)

art-dsit:
Claude wrote this.

Unfortunately it doesn't work - see https://forum.proxmox.com/threads/maximum-file-upload-size-for-qemu-agent-file-write.166200/


art-dsit:
We don't need this if we have the ISO fast path

PR #14, opened 2025-10-14
Change VM environment names

pgiav:
# Better VM Naming in Proxmox Provider

## Current Issue Explanation

The current implementation uses dynamically generated IDs (`vm_<id>`) to register VMs in Inspect. This approach is impractical for multiple reasons- all revolve around the id part not being easily predictable, especially when you run multiple epochs of a sample. 

If, for instance, you want to run `sandbox().exec()` on a VM that is not the `default` sandbox, you'd have to know the VM ID beforehand and specify it like so: `sandbox("vm_101").exec()`. The same applies if you [want to copy files into the VM](https://inspect.aisi.org.uk/sandboxing.html#files).   

## The Patch
This patch enables setting VM names instead of arbitrary IDs. It uses whatever you define in the `VmConfig`'s optional `name=` parameter, and falls back to the `vm_<id>` convention otherwise.  

## Example Usage

Define VMs with meaningful names:

```python
...
VmConfig(
    vm_source_config=VmSourceConfig(
        built_in="ubuntu24.04"
    ),
    name="romeo",
)
```

So then you can access the environment from within a tool with `sandbox("romeo")`, or write a file to it via `"romeo:/root/flag.txt": "flag.txt"`.

The important thing is that this actually scales with multiple epochs- so there's no clash when Inspect tries to write the file into VMs belonging to different epochs with the same `romeo` name. 

> I have verified this when playing around with `inspect_cyber`, but I haven't managed to write a test that shows this yet... Can do that if needed.

The other important thing is that the first sandbox VM gets renamed to `default`, so one shouldn't try naming and accessing that via some custom name. I can put that in the `README.md`.

## Technical Details

- Names don't clash in Proxmox when running multiple epochs
- Falls back to `vm_<id>` if no name is provided, so it's backward compatible 
- Simplifies complex scoring scripts that need to target specific VMs

## Questions for Review

Should I add a warning or error check for when two VMs have been defined with the same name in a single sandbox configuration? I don't know why anyone would do this, but currently the dictionary logic will simply overwrite whichever VM is defined first as it uses the name as a key.

Do you need those tests for multiple epochs or is that something that technically belongs to the Inspect part of things so we shouldn't test for it here?

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ It is recommended that you set the `name=` parameter for your defined VMs. This 
 - It will be displayed in the Proxmox web interface
 - It will be the identifier you use to reference the VM in Inspect (e.g., `sandbox("vm_name")`)
 
-You should avoid setting the same name for multiple VMs as this will cause conflicts; later VMs with the same name will overwrite earlier ones. If you omit the name parameter, the VM will be registered in Inspect using its dynamically-generated ID, as `vm_<id>`.
+You should avoid setting the same name for multiple VMs as this will cause conflicts in how Inspect references your VMs; later VMs with the same name will overwrite earlier ones in the sandbox name mapping. While both VMs would still be created in Proxmox, only the last one would be accessible through its name in Inspect. If you omit the name parameter, the VM will be registered in Inspect using its dynamically-generated ID, as `vm_<id>`.
 
 > Note: The (first) sandbox VM is automatically named `default` internally, so you can always access it with `sandbox("default")`, regardless of any custom name you might set for it.
 


tests/proxmoxsandboxtest/test_proxmox_sandbox_environment_e2e.py
```diff
@@ -241,6 +241,64 @@ async def test_connect(proxmox_sandbox_environment: ProxmoxSandboxEnvironment) -
...
+        # We can access this one via its name now
+        sandbox = envs_dict["other-vm"]
+        uname_result = await sandbox.exec(
+            [
+                "uname",
+                "-a",
+            ]
+        )
+        assert uname_result.success, f"Failed to run uname: {uname_result=}"
+        assert "Ubuntu" in uname_result.stdout, (
```
art-dsit: It's a shame Ubuntu is the only built-in at the moment, otherwise you could use the test to differentiate two different OSs and check the mapping. With the new Proxmox 9 release I'm going to add support for direct qcow uploads which should allow Debian and Kali

art-dsit:
> Should I add a warning or error check for when two VMs have been defined with the same name in a single sandbox configuration? I don't know why anyone would do this, but currently the dictionary logic will simply overwrite whichever VM is defined first as it uses the name as a key.

I think it's OK to just document this as something to avoid doing

> Do you need those tests for multiple epochs or is that something that technically belongs to the Inspect part of things so we shouldn't test for it here?

I think a test for that would be great but if you leave it as a TODO in the test file that's OK by me

art-dsit (approved):

PR #22, opened 2025-11-21
Multi-proxmox usage

sinman-aisi:
This commit allows usage of multiple proxmox servers to run evals concurrently. 

The current implementation is a queue-based system where each sample is run in a separate proxmox instance. There is an abstract base class that would allow implementation of different setups if desired. 

Config of the proxmox servers is in a JSON config file. This maintains backwards compatibility to using a single proxmox instance with env vars.

### DIFF
--- a/src/proxmoxsandbox/_proxmox_sandbox_environment.py
+++ b/src/proxmoxsandbox/_proxmox_sandbox_environment.py
@@ -236,7 +236,7 @@ class ProxmoxSandboxEnvironment(SandboxEnvironment):
                 async_proxmox_api=async_proxmox_api, config=config
             )
 
-            async with concurrency("proxmox", 2):
+            async with concurrency(f"proxmox-{instance.instance_id}", 1):
                 vm_configs_with_ids, sdn_zone_id = await infra_commands.create_sdn_and_vms(
                     proxmox_ids_start,
                     sdn_config=config.sdn_config,


src/proxmoxsandbox/_proxmox_sandbox_environment.py
```diff
@@ -51,6 +58,9 @@ class ProxmoxSandboxEnvironment(SandboxEnvironment):
     vm_id: int
     all_vm_ids: Tuple[int, ...]
     sdn_zone_id: str | None
+    # Multi-instance pool fields
+    instance: ProxmoxInstanceConfig | None
+    pool_id: str | None
```
willpayne23: Do we use these anywhere?

src/proxmoxsandbox/_proxmox_sandbox_environment.py
```diff
@@ -182,71 +220,92 @@ async def sample_init(
...
+            await ProxmoxSandboxEnvironment.ensure_vms(
+                async_proxmox_api=async_proxmox_api, config=config
             )
 
-        sandboxes: Dict[str, SandboxEnvironment] = {}
-
-        vm_ids = tuple(
-            vm_configs_with_id[0] for vm_configs_with_id in vm_configs_with_ids
-        )
+            async with concurrency("proxmox", 2):
```
willpayne23: Why is this hard set to 2?

src/proxmoxsandbox/_proxmox_sandbox_environment.py
```diff
@@ -268,20 +327,32 @@ async def sample_cleanup(
...
+
+        try:
+            if not interrupted and any_vm_sandbox_environment is not None:
                 async with concurrency("proxmox", 1):
                     await any_vm_sandbox_environment.infra_commands.delete_sdn_and_vms(
                         sdn_zone_id=any_vm_sandbox_environment.sdn_zone_id,
                         vm_ids=any_vm_sandbox_environment.all_vm_ids,
                     )
+        finally:
+            # RELEASE instance back to pool (even if cleanup failed)
```
willpayne23: Maybe we don't want to do this if cleanup failed (again something for me to look into)

sinman-aisi: Yeh, I agree, was just looking at this before sending for review. I'm slightly worried about just losing a bunch of instances, but that's obviously better than crashing (which is what will happen if the cleanup has failed).  Unless it exhausts all the proxmoxes, in which case things will hang forever. 

I think I'll change this to not return to the pool for now?

src/proxmoxsandbox/_proxmox_sandbox_environment.py
```diff
@@ -155,20 +169,44 @@ def config_files(cls) -> List[str]:
...
+
+        Proper per-pool concurrency control would require changes to Inspect's
+        core architecture, which is not designed to handle heterogeneous
+        limited sandbox environments within a single task.
+
+        Returns:
+            Total number of Proxmox instances, or 1 for legacy single-instance mode.
+        """
+        try:
+            instances = _load_instances_from_env_or_file()
```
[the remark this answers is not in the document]
sinman-aisi: Good catch, thank you! Have updated.

src/proxmoxsandbox/schema.py
```diff
@@ -193,31 +195,102 @@ class VmConfig(BaseModel, frozen=True):
...
+            PROXMOX_CONFIG_FILE or defaults to "default" for legacy single-instance mode)
         sdn_config: Software-defined networking configuration
-            "auto": Create a simple SDN with a single subnet.  The IP addresses will not
-                be predictable as it depends on what subnets already exist.
-            None: No SDN will be created. VMs can reference existing VNETs in Proxmox
-                by their aliases. This is an advanced feature and not recommended for
-                normal use.
-            SdnConfig: Custom SDN configuration
         vms_config: Configurations for virtual machines
+        host: (Legacy) The hostname or IP address of the Proxmox server
```
art-dsit: I don't agree that this is "legacy" - I think users who are new to this would do well to stick with a single proxmox instance

src/proxmoxsandbox/schema.py
```diff
@@ -193,31 +195,102 @@ class VmConfig(BaseModel, frozen=True):
     ] = "l26"
 
 
-class ProxmoxSandboxEnvironmentConfig(BaseModel, frozen=True):
+class ProxmoxInstanceConfig(BaseModel, frozen=True):
```
art-dsit: I think I would rather this had the host, port, user, user_realm, node, and verify_tls (as before) as its own class (`ProxmoxInstanceConnectionConfig`) ? then both this class and `ProxmoxSandboxEnvironmentConfig` can have it as a field

tests/proxmoxsandboxtest/test_config_loading.py
```diff
@@ -0,0 +1,393 @@
...
+from proxmoxsandbox.schema import (
+    ProxmoxInstanceConfig,
+    ProxmoxSandboxEnvironmentConfig,
+    _load_instances_from_env_or_file,
+)
+
+
+def test_load_instances_from_file():
+    """Test loading instances from PROXMOX_CONFIG_FILE."""
+    config_data = {
```
art-dsit: nit - IMHO this should just actually be a static file in the tests folder rather than writing it out each time

CLAUDE.md
```diff
@@ -0,0 +1,398 @@
+# Claude Code Context for inspect-proxmox-sandbox
```
art-dsit: In general I would rather the human docs be improved; this document is duplicative of the README. Maybe some of it could be moved into the README?

poetry.lock
art-dsit: Sir, this is a uv project

PR #24, opened 2025-11-21
Add multi proxmox cleanup

willpayne23:
Stacked on #22 

This PR adds CLI and task cleanup with awareness for multi proxmox setups at task cleanup time and at cli cleanup it iterates through all proxmox instances configured and cleans them up.

Currently the cli cleanup just does a confirmation dialogue for each proxmox configured (so for 50 proxmoxes you would need to confirm each cleanup. It would require a bit of a restructure of the logic to have only one confirmation currently the listing of which instances would be destroyed is in infra commands.

This would probably involve first gathering via an infra_command function (looping over all proxmox instances) then printing the list of VMs and networks scheduled to be destoyed (across all instances) then looping through and deleting them after aquiring confirmation.

### DIFF
--- a/tests/proxmoxsandboxtest/test_cli_cleanup.py
+++ b/tests/proxmoxsandboxtest/test_cli_cleanup.py
@@ -1,0 +1,294 @@
+"""Tests for the cli_cleanup function."""
+
+import json
+import os
+import tempfile
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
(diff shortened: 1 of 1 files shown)


sinman-aisi (commented):
CLI command seems good - did you test it?  Would be good to test if easy. And useful to have that function this week. 

As I commented I'm not sure on the Task changes - seems dangerous to start cleaning up stuff unless we're sure we'll only ever have one Task.

PR #34, opened 2025-12-03
rejig static IP

art-dsit:
Changes to your PR for discussion

* I tried to be consistent calling it "IPAM Mapping" everywhere, but might have missed one or two places
* During sample setup, we save the list of ipam_mappings so we can tear them down afterwards
* task_cleanup now actually works, so Ctrl-C cleans up properly
* cli_cleanup deletes all IPAM mappings in relevant zones

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -62,7 +62,7 @@ uv run ruff format .
 manually:
 
 ```bash
-mypy
+uv run mypy
 ```
 
 ## Design Notes
--- a/README.md
+++ b/README.md
@@ -237,6 +237,8 @@ sandbox = SandboxEnvironmentSpec(
 )
 ```
 
+Static IP address assignment is *not* supported with this feature.
+
 ## Using OVA files
 
 Proxmox supports OVA import but not OVA export. It is possible to extract the disk images of VMs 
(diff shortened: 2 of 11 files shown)


pgiav (approved):
This looks good to me. The only snag I see is that edge case of combining behaviours like referencing a pre-existing VNET and trying to set a static IP. Currently, AFAICT, this will just fail silently (we `return []` in `infra_commands/create_ipam_mappings`). 

I tried to document that issue in the code: 

```python
    async def create_sdn(
        self, proxmox_ids_start: str, sdn_config: SdnConfigType
    ) -> Tuple[Optional[str], VnetAliases]:
        if sdn_config is None:
            # When sdn_config is None, we should still fetch existing VNETs
            # so we can use them for VM network interfaces
            all_vnets = await self.read_all_vnets()
            vnet_aliases = []
            for vnet in all_vnets:
                if "vnet" in vnet and "alias" in vnet:
                    vnet_aliases.append((vnet["vnet"], vnet["alias"]))
            # This returns None because we cannot make any guarantees about
            # existing VNETs and what zones they belong to. We _could_ read
            # the zone attribute returned by `read_all_vnets`, but different
            # VNETs could belong to different zones.
            return None, vnet_aliases
```

Maybe that whole implementation of referencing stuff not defined by the Proxmox Provider was muddying the waters, for which I take full blame. 

Nevertheless, we should add a `README.md` entry at minimum, and maybe fail early in the code if this happens?

PR #40, opened 2026-02-26
feat: allow_domains egress filtering for sandbox VMs

RoshDSIT:
## Summary

Adds an `allow_domains` option to `SdnConfig` that restricts sandbox VM internet egress to a specified domain allowlist. Enforcement is out-of-band (gateway VM outside the sandbox), so root inside the sandbox cannot bypass it.

**How it works:**
- A gateway VM (Debian 13, cloned per eval sample from a one-time-built template `inspect-gw`) sits between sandbox VMs and the internet
- dnsmasq on the gateway intercepts all DNS from the sandbox and only forwards queries for allowed domains (returning SERVFAIL for everything else)
- nftables on the gateway drops all forwarded traffic except to IPs in an `allowed_ips` set
- Allowed domain IPs are pre-resolved at provision time and added to the set

**Usage:**
```python
sdn_config=SdnConfig(
    vnet_configs=(...),
    use_pve_ipam_dnsnmasq=True,
    allow_domains=("pypi.org", "files.pythonhosted.org"),
)
```

**Costs:** gateway template built once (~5–10 min). Per-eval overhead ~30–60 s for clone boot.

## Key implementation details

- **ARP conflict fix:** Proxmox's `SubnetConfig.gateway` field couples the bridge device IP and DHCP option-3 in a single field — setting `gateway=.2` causes both the gateway VM and the Proxmox bridge to claim `.2`. Resolved by injecting `nud permanent` ARP entries on sandbox VMs after boot, which take unconditional precedence over dynamic ARP replies.
- **Static IP persistence:** gateway VM's sandbox-facing NIC is configured via a systemd-networkd `.network` file (not `ip addr add`), so the address survives DHCP renewals.
- **DNS interception:** nftables prerouting redirects all UDP/TCP port 53 from the sandbox CIDR to the gateway's dnsmasq, preventing DNS bypass via alternative resolvers.
- **Pre-resolve:** allowed domain IPs are resolved at provision time using `python3-dnspython` querying 8.8.8.8 directly (matching dnsmasq's upstream) and seeded into the nftables set.
- **Cleanup idempotency:** SDN zone deletion now tolerates zones that were already deleted out-of-band (e.g. via the Proxmox UI), so gateway VM clones are not left orphaned when cleanup is re-run.
- **IPv6 SLAAC blocked:** built-in VM cloud-init sets `accept-ra: false` (not just `dhcp6: false`); the gateway's `table inet` forward chain policy drop covers IPv6 forwarded through it.
- **DoT blocked:** nftables forward chain drops `ip dport 853` before the `@allowed_ips accept` rule, closing the DNS-over-TLS bypass path.

## Known limitations (documented in schema docstring)

- Filtering is IP-level, not URL-level — all ports to allowed-domain IPs are permitted
- Only apex domain IPs are pre-seeded; subdomains require explicit listing (e.g. `ftp.gnu.org` when only `gnu.org` is listed will be dropped). Note: domains sharing CDN anycast IPs (e.g. two domains both behind Fastly) will not exhibit this gap — allowing the apex inadvertently allows co-hosted subdomains. A future improvement is to enable dnsmasq `nftset=` support for dynamic IP seeding.
- IPv6 SLAAC protection via `accept-ra: false` applies only to built-in VM templates; custom VM sources (OVA, `existing_vm_template_tag`) must configure this independently.
- `allow_domains` currently requires exactly one `vnet_config` — multi-vnet support is deferred to a follow-up PR.

## Test plan

- [x] `test_allow_domains` e2e test: allowed domain (cloudflare.com) reachable, blocked domain (google.com) fails, direct IP (8.8.8.8) fails
- [x] `test_allow_domains_manual.py` — first 4 manual tests passing (requires Proxmox env):
  - `test_proxmox_host_ip_is_blocked`: `curl` to the Proxmox host's RFC1918 IP fails (never in nftables set)
  - `test_subdomain_blocked_when_only_apex_listed`: `ftp.gnu.org` (IP `.20`) is unreachable when only `gnu.org` (IP `.116`) is listed
  - `test_subdomain_allowed_when_listed_explicitly`: `ftp.gnu.org` is reachable when listed explicitly alongside `gnu.org`
  - `test_all_ports_open_to_allowed_ips`: TCP port 443 to cloudflare.com reachable; google.com (not listed) unreachable
- [ ] Two new manual tests added for IPv6/DoT — need verification against live Proxmox:
  - `test_ipv6_slaac_blocked`: no global-scope IPv6 address on sandbox VM (`accept-ra: false` effective); IPv6 curl to allowed domain fails
  - `test_dns_over_tls_port_853_blocked`: port 443 to cloudflare IP reachable (IP in set), port 853 to same IP blocked (drop rule fires first)
- [ ] If upgrading an existing deployment: delete the `inspect-gw` template VM and let it rebuild to pick up `python3-dnspython`

Original ticket: [EVA-179](https://linear.app/aisi/issue/EVA-179/support-partial-internet-allowlist-in-proxmox-provider)

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,10 +266,11 @@ sdn_config=SdnConfig(
 )
 ```
 
-Subdomains must be listed explicitly: `"debian.org"` does **not** cover
-`"deb.debian.org"` or `"security.debian.org"` — list each subdomain you need.
-See the `allow_domains` docstring in `schema.py` for full constraints and
-caveats.
+`"debian.org"` covers the apex and all subdomains (`deb.debian.org`,
+`security.debian.org`, etc.) — no wildcard prefix needed.  Subdomain IPs
+are injected dynamically on first DNS query via dnsmasq `nftset=`.
+See the `allow_domains` docstring in `schema.py` for full constraints
+and caveats.
 
 **Costs:** the gateway template (`inspect-gw`) is built once (~5–10 min).  Per-eval
 overhead is ~30–60 s for the gateway clone to boot.
--- a/README.md
+++ b/README.md
@@ -266,8 +266,10 @@ sdn_config=SdnConfig(
 )
 ```
 
-`"debian.org"` covers the apex and all subdomains — no `*.` prefix needed.  See the
-`allow_domains` docstring in `schema.py` for full constraints and caveats.
+Subdomains must be listed explicitly: `"debian.org"` does **not** cover
+`"deb.debian.org"` or `"security.debian.org"` — list each subdomain you need.
+See the `allow_domains` docstring in `schema.py` for full constraints and
+caveats.
 
 **Costs:** the gateway template (`inspect-gw`) is built once (~5–10 min).  Per-eval
 overhead is ~30–60 s for the gateway clone to boot.
--- a/README.md
+++ b/README.md
@@ -239,6 +239,39 @@ sandbox = SandboxEnvironmentSpec(
 
 Static IP address assignment is *not* supported with this feature.
 
+### Egress filtering with allow_domains
+
+Set `allow_domains` on `SdnConfig` to restrict sandbox internet access to specific domains.
+A gateway VM (cloned per eval sample) enforces the allowlist via dnsmasq + nftables.
+Because the gateway is outside the sandbox, root inside the sandbox cannot bypass it.
+
+```python
+sdn_config=SdnConfig(
+    vnet_configs=(
+        VnetConfig(
+            alias="sandbox-net",
+            subnets=(
+                SubnetConfig(
+                    cidr=ip_network("10.10.0.0/24"),
+                    gateway=ip_address("10.10.0.1"),  # overridden internally
+                    snat=True,                        # overridden internally
+                    dhcp_ranges=(DhcpRange(start=ip_address("10.10.0.10"),
+                                          end=ip_address("10.10.0.100")),),
+                ),
+            ),
+        ),
+    ),
+    use_pve_ipam_dnsnmasq=True,
+    allow_domains=("pypi.org", "files.pythonhosted.org"),
+)
+```
+
+`"debian.org"` covers the apex and all subdomains — no `*.` prefix needed.  See the
+`allow_domains` docstring in `schema.py` for full constraints and caveats.
+
+**Costs:** the gateway template (`inspect-gw`) is built once (~5–10 min).  Per-eval
+overhead is ~30–60 s for the gateway clone to boot.
+
 ## Using OVA files
 
 Proxmox supports OVA import but not OVA export. It is possible to extract the disk images of VMs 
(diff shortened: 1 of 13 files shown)


src/proxmoxsandbox/schema.py
```diff
@@ -69,10 +70,96 @@ class SdnConfig(BaseModel, frozen=True):
...
+            When allow_domains is non-empty:
+            - snat on sandbox subnets is set to False (the gateway VM does NAT)
+            - An extra internal VNet is created for the gateway VM's external interface
+            - Sandbox VMs learn the gateway as their default router via DHCP
+            - The gateway IP in SubnetConfig is overridden to network-address+2; any
+              value you specify there will be silently replaced.
+
+            Constraints (validated at construction time):
+            - use_pve_ipam_dnsnmasq must be True
+            - exactly one vnet_config must be provided
```
sinman-aisi: This won't work for our purposes - our evals can have multiple VNet configs, such as when we're representing a range with multiple different networks. Is this fixable?

src/proxmoxsandbox/_impl/infra_commands.py
```diff
@@ -22,11 +26,183 @@
...
+
+    listen-address is set to the gateway's sandbox-facing IP only.
+    bind-interfaces is required alongside listen-address: without it, dnsmasq
+    binds to 0.0.0.0:53 regardless of listen-address (listen-address only
+    affects filtering, not binding).  bind-interfaces makes dnsmasq bind to
+    exactly the specified IP, so the restart doesn't race with the boot-time
+    dnsmasq instance releasing 0.0.0.0:53.
+
+    nftset= is intentionally omitted: Debian 13's dnsmasq package may not be
+    compiled with --enable-nftset.  The nftables allowed_ips set is populated
```
artbristol: either it is or isn't compiled with `--enable-nftset` and it should be easy to check, so this comment seems odd

artbristol (commented):
This is great, and fixes a big functionality gap. I am back in work on Monday, if you want to discuss?

README.md
```diff
@@ -295,18 +328,23 @@ SDN zones have the pattern `[3 letters from eval task name][random 3 digits][z]`
...
+created by a particular eval. Therefore the cleanup process for `inspect sandbox cleanup proxmox`
 will delete:
 
-- all VMs tagged `inspect` 
+- all VMs tagged `inspect`
 - any SDN zones created with names matching the pattern above.
 
+If an eval is interrupted (e.g. Ctrl+C), per-sample cleanup is skipped and VMs are left
+running. This includes any gateway VM clones created for `allow_domains` evals. Run
+`inspect sandbox cleanup proxmox` to collect them.
```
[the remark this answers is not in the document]
RoshDSIT: Have confirmed - on single Ctrl+C: Textual cancels the worker → asyncio.CancelledError propagates → `sandboxenv_context` catches it, sets `interrupted=True`, re-raises → `sample_cleanup` runs but skips VM deletion → eval log marked "cancelled", but VMs stay running. Cleanup reuqired with `inspect sandbox cleanup proxmox`.

pgiav:
Any updates on this?

RoshDSIT:
[Claude wrote this - not incorrect but didn't ask it to!]

Hey @pgiav — sorry for the delay, been working through this. Multi-vnet support is in progress and will land on this same PR. Quick update on where we are:

**Done today:**
- IPv6 SLAAC blocking and DNS-over-TLS (port 853) blocking both validated on live Proxmox — passing
- Fixed a bug in the nftables rules (wrong syntax for the port 853 drop rule)
- Added UDP port 853 blocking too (closes DNS-over-QUIC bypass)
- Added domain string validation to prevent malformed input

**Still to do on this PR:**
- Multi-vnet `allow_domains` — the gateway VM gets N+1 NICs (one per sandbox VNet + external). Architecture is straightforward, working on it now.

**Two things to flag / discuss:**

1. **CDN IP rotation** — right now we pre-resolve allowed domain IPs once at provision time and seed them into nftables. If a domain rotates its IPs during the eval (CDN load balancing, GeoDNS), the new IPs aren't in the set and traffic gets dropped. This affects any domain behind a CDN — PyPI, GitHub, npm, etc. The fix is to enable dnsmasq's `nftset=` directive alongside pre-resolve (boot-time seeding + runtime additions). Probably a follow-up — flagging so we're aware of the limitation.

2. **Hardcoded 8.8.8.8 as upstream DNS** — dnsmasq forwards allowed domain queries to 8.8.8.8 directly. Environments that block outbound DNS to 8.8.8.8 would silently fail. Worth making configurable?

Will update when multi-vnet is ready for testing.

art-dsit:
reimplemented in #59

PR #43, opened 2026-03-05
Allow storage location to be configured

art-dsit:
This will allow use on a Proxmox instance that was created [on top of Debian](https://pve.proxmox.com/wiki/Install_Proxmox_VE_on_Debian_13_Trixie) rather than installed using the ISO


pgiav (approved):
LGTM

PR #47, opened 2026-03-12
Q4 scaling merge in main

art-dsit:
Another go at #44 ; tests passing now


pgiav:
> Hmm, won't this interfere with other samples that are in progress?

I think `task_cleanup` only runs when all samples are done, so there shouldn't be anything else in progress at that point. However, I guess if you were re-using the Proxmox instance for multiple Inspect processes, then those other ones would be deleted

pgiav:
I ran a test on our suite and no issues came up 👍

PR #49, opened 2026-03-19
Allow Docker pull through cache to be specified

art-dsit:
So you can still build your proxmox image even if DockerHub is having a lie down


bmillwood-aisi (approved):
I noticed one gotcha but up to you whether anything needs to be done.

bmillwood-aisi:
Presumably you could also just do something like:
```
docker pull alternative/repo/debian:bookworm-slim
docker tag alternative/repo/debian:bookworm-slim debian:bookworm-slim
```

and not modify the Dockerfile?

(and actually this seems kinda simpler / preferable to me?)

PR #55, opened 2026-03-23
Rationalize task cleanup

art-dsit:
Now that object creation is better factored (in #50), we can store the known resources (VM IDs, SDNs, IPAM mappings) in normal fields instead of ContextVars, which were confusing and unnecessary. The only tricky part is finding the relevant `InfraCommands` for a given sample, which is stored in a map in a class-level variable now.


pgiav (approved):
lgtm, provided that tests are passing

PR #57, opened 2026-03-30
Allow vend script to be generated standalone

art-dsit:
Add `VEND_VEND_SCRIPT` option to build_proxmox_auto. 

If you're using a pre-built qcow file for `proxmox-auto`, but you want a fresh vend script, you can set VEND_VEND_SCRIPT=1 to avoid the delay of building proxmox each time.

It's a big diff but it's only moving code around.


pgiav (approved):
lgtm, thanks!

PR #63, opened 2026-04-30
Fix edge-case in network cleanups.

pgiav:
It's possible to hook up a VM to a predefined VNET/Zone in Proxmox that is **not** created by the provider during eval-build-time. However, there's a long-standing edge case that makes it so that if a VM is attached to a predefined VNET, that VNET and its zone can be wiped by the provider:

- If the sample runs to completion, the cleanup is fine (the predefined VNET persists).
- If the sample is `ctrl+c`'d, the cleanup is fine (the predefined VNET persists).
- If, however, a dirty sample is interrupted and wiped manually via `inspect sandbox cleanup proxmox`, the VNET and parent Zone are wiped from the instance. The condition is that a provider-defined VM (i.e. one that has the `inspect` tag) is cloned and attached to that predefined VNET. 

This PR constrains `cleanup_no_id` to only delete zones whose names match the provider's ephemeral-zone naming convention (already enforced at zone creation). Pre-existing user zones, and the static `inspvm*` SDN, are skipped, even when an inspect-tagged VM happens to be plugged into one of their VNETs. The pre-check warning is also filtered so user VNETs no longer trip a misleading "leftover VNETs" message.

---

- [X] Tested the new tests
- [X] All tests fine

### DIFF
--- a/src/proxmoxsandbox/_impl/infra_commands.py
+++ b/src/proxmoxsandbox/_impl/infra_commands.py
@@ -278,14 +278,27 @@ class InfraCommands(abc.ABC):
                         noticed_vnets.add(bridge)
                 noticed_vms.append(vm)
 
-        zones_to_delete = await self.find_all_zones(noticed_vnets)
+        # Only zones matching the provider's ephemeral-zone naming convention
+        # are eligible for sweep — `create_sdn` enforces this convention at
+        # creation (sdn_commands.create_sdn). Pre-existing user zones
+        # (referenced via sdn_config=None) and the intentionally-permanent
+        # static `inspvm*` SDN do not match, and must not be deleted even
+        # when an `inspect`-tagged VM is plugged into one of their VNETs.
+        def _is_provider_zone(zone_id: str) -> bool:
+            return bool(re.match(ZONE_REGEX, zone_id))
 
-        # We probably already have all the SDN zones already.
-        # But in case there were no VMs in a particular SDN zone
-        # (which can happen if the task setup failed)
-        # we need to check for orphans.
+        # Zones that VMs we created are plugged into. Filtered to provider
+        # zones only, so attaching a sandbox VM to a user-pre-existing VNET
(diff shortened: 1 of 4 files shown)


pgiav:
To illustrate the edge case:

```bash
# on q4_scaling HEAD:
ubuntu@bobcat ~/sandbox pgiav/staging-cleanup-fix*
❯ git checkout 620cc
M       uv.lock
Note: switching to '620cc'.

...
HEAD is now at 620cced Remove internal exclude-newer from uv.lock

ubuntu@bobcat ~/sandbox heads/q4_scaling*
❯ uv run inspect sandbox cleanup proxmox
┌───────┬─────────┐
│ VM ID │ VM Name │
├───────┼─────────┤
│ 124   │ zeeknsm │
└───────┴─────────┘
┌──────────┐
│ Zone ID  │
├──────────┤
│ a254c5f5 │
└──────────┘
Are you sure you want to delete ALL the above resources? [y/n]: y

...
# after my changes; no more Zone ID 
ubuntu@bobcat ~/sandbox pgiav/staging-cleanup-fix*
❯ uv run inspect sandbox cleanup proxmox
┌───────┬─────────┐
│ VM ID │ VM Name │
├───────┼─────────┤
│ 124   │ zeeknsm │
└───────┴─────────┘
┌─────────┐
│ Zone ID │
├─────────┤
│ (none)  │
└─────────┘
Are you sure you want to delete ALL the above resources? [y/n]: y
❯
```

sinman-aisi:
Sorry for ignorance - why do we have pre-defined zones at all, rather than setting stuff up from the sandbox.yaml? Is there some stuff that can't be set up from sandbox?

pgiav:
> Sorry for ignorance - why do we have pre-defined zones at all, rather than setting stuff up from the sandbox.yaml? Is there some stuff that can't be set up from sandbox?

There's monitoring-related use cases that require this. Can give you more details on slack.

art-dsit (approved):

PR #71, opened 2026-05-11
feat: ISO hot-plug fast path for write_file (Linux)

art-dsit:
## Summary

Sidesteps the QGA `guest-file-write` cap (~60 KiB per call through Proxmox's Agent.pm wrapper) for large `write_file` calls on Linux guests. Packs the payload into an ISO9660 image, uploads to Proxmox storage, swaps it into a **dedicated `sata5` CD-ROM slot cold-added at VM clone time**, then has the guest agent mount + `cp` + unmount.

Linux only for now. Falls back unconditionally to the existing chunked-QGA path on any failure, so this is a pure optimisation — never a regression.

## Measurements

Live, against a freshly cloned ubuntu24.04 sandbox on a NitroVirt Proxmox host (m8i):

| Size | ISO path (this PR) | QGA chunked (current main, c=1) |
|---|---|---|
| 2 MiB | 2.6 s | 36 s |
| 8 MiB | 2.6 s | — (not re-measured; scales linearly with chunks) |
| 20 MiB | 2.7 s | — |
| 56 MiB | 2.6–2.9 s | — (PR #70 reports ~10 min at c=8) |

Per-call overhead is now ~2.6 s and nearly flat with size. Breakdown for a 2 MiB write:

```
total=2.60s build=0.00s upload=1.04s attach_copy=1.52s detach=0.02s delete=0.02s
```

Of which ~1.0 s is the curl POST to Proxmox storage and ~1.5 s is QGA exec + status-polling round-trip overhead. The actual mount/cp/umount inside the guest is negligible.

## Design notes

- **Slot must be cold-added, not hot-added.** Empirically verified on this Proxmox build (both i440fx and q35 machine types): a `POST /config` adding a previously-absent `sataN` or `ideN` slot to a running VM is silently dropped — the field doesn't even persist in the GET-back config. The slot must exist in the VM's boot-time config for QEMU to enumerate the controller. Cold-add of `sata5: none,media=cdrom` in `clone_vm_and_start`'s `other_updates` block works cleanly; runtime media-change on that slot then maps to QEMU's monitor `change` command which the guest sees as a media-change event.
- **Why `sata5`, not `ide3`.** IDE has only 4 slots (`ide0`–`ide3`); SATA has 6 (`sata0`–`sata5`). Picking the highest sata slot minimises collisions with `existing_vm_template_tag` templates that happen to populate optical drives. SATA also auto-instantiates an AHCI controller when referenced; we don't need to flip `machine` to `q35` for it to work.
- **Write serialisation is a per-VM lock that lives on the `ProxmoxSandboxEnvironment` (one env per VM), not a module-level dict keyed on `vm_id`.** Concurrent writes to a VM share the single `sata5` slot and must not overlap their media-changes — but `vm_id` is only unique *within* one Proxmox host (our hosts are unclustered singletons, each vending IDs from ~100 via `/cluster/nextid`), so a module-global dict keyed on the bare integer would have falsely serialised VM 100 on host A against VM 100 on host B. Anchoring the lock on the env object scopes it to exactly one physical VM. The fast-path block also re-checks the disable flag *after* acquiring the lock, so a write that queues behind a failing one falls straight through to chunked QGA.
- **Per-VM fast-path-failure memoisation.** The first failure on a VM sets `env._iso_fast_path_disabled` and emits a `WARNING`; subsequent large writes on that VM skip straight to chunked QGA rather than re-paying ~3 s of ISO build+upload+attach before falling back. Other VMs are unaffected.
- **Size-gated at 128 KiB** on Linux; smaller writes and all Windows traffic fall through to the existing chunked-QGA path. Benchmarking (live ubuntu24.04, both paths driven end-to-end through `write_file`) found ISO actually wins down to ~8 KiB — it collapses to one in-guest exec, beating even QGA's separate mkdir+write round-trips — so 128 KiB is not a true crossover. It's set deliberately above it so trivially-small writes stay on the simpler chunked path (no storage upload, smaller failure surface) and keep that fallback exercised on the normal route. Measured speedups in the 64 KiB–1 MiB band were 7–13×.
- **First-call "Can't open blockdev" race.** On a freshly-booted VM, the kernel sometimes refuses every `open()` on `/dev/sr*` after the first media-change — persistent for the full mount-retry budget, dmesg shows repeated "Can't open blockdev" errors. Detaching and re-attaching emits a fresh media-change event which the kernel then handles cleanly. One host-side retry on `_copy_in_guest` failure handles this case; in-guest loop budget tightened to ~3s so the retry triggers fast rather than hanging on the failed first attempt for 30s.
- **Perf wins beyond the QGA-cap workaround itself.** Two avoidable costs in the original draft:
  1. `task_wrapper.do_action_and_wait_for_tasks` has a hardcoded 2 s mandatory `asyncio.sleep` before polling. For an ephemeral random-named ISO, the curl POST already returns when the file is on disk; bypass the task-wait layer entirely. Saved ~2 s.
  2. `write_file` did `mkdir -p` of the target's parent via `env.exec()` — a full QGA round-trip (~2.3 s for a no-op mkdir, dominated by exec wrapper overhead, not the mkdir itself). The in-guest ISO script does its own `mkdir -p`, so the outer call is pure overhead on the fast path. Saved ~2.3 s.

## Still to do before un-drafting

- [ ] **Windows support.** Mechanism transfers (ISO9660 readable everywhere; `change` monitor command works the same). Differences: drive-letter discovery instead of `/dev/sr0`; no mount step (Windows auto-mounts CD media). Needs verification against an actual Windows sandbox template.
- [x] ~~Smarter error handling / fallback~~ — host-side single re-attach retry on `_copy_in_guest` failure for the "Can't open blockdev" race, plus per-VM memoisation that disables the fast path after a persistent failure so we don't re-pay ISO overhead on every subsequent large write.
- [ ] Decision on whether "storage full" / "permission denied" should fall back to QGA at all, or hard-fail loudly. Current behaviour: any exception → fallback.

## Out of scope

- Orphan ISO sweeper on hard process kill — not worth the complexity for the prototype; normal exits clean up via `finally`.
- Batching multiple `write_file` calls within a sample into one ISO — would be another big win for setup-heavy samples, but a separate change.

## Test plan

- [x] Type checks pass (`uv run mypy`)
- [x] Unit tests pass (`tests/proxmoxsandboxtest/test_iso_write.py`) — covering ISO round-trip, the per-env write lock (incl. a regression that two envs sharing a `vm_id` get distinct locks), the `sata5` slot constant, and fast-path memoisation
- [x] e2e `test_write_file_large[linux]` (56 MiB write) passes — exercises the ISO path end-to-end
- [x] mockllm probing eval `test_eval.py::test_iso_write_fast_path_via_eval` passes (~39 s, live host) — writes ~5 MiB, sha-verifies in-guest, snapshots guest dmesg/`/dev/sr*`/AHCI state before+after, and asserts the `iso_write vm=…` log line appears (fast path ran, not the fallback) and the "fast path disabled" warning does not
- [ ] Force-failure path: when the first attach trips "Can't open blockdev", the host-side re-attach retry should kick in. Code path is wired up but I haven't reliably reproduced the original failure since adding the shorter loop budget, so this isn't actively verified end-to-end.
- [ ] Manual: write same sizes to a Windows sandbox (once Windows path lands)


sinman-aisi:
@art-dsit can we merge this? Even without windows would be 99% of the use cases. Saves about 18 minutes per transatomic run, which is very valuable.

art-dsit:
> @art-dsit can we merge this? Even without windows would be 99% of the use cases. Saves about 18 minutes per transatomic run, which is very valuable.

Happy to as long as everything is working OK for you?

PR #72, opened 2026-05-28
Block sandbox VMs from reaching the Proxmox host

art-dsit:
## Summary

Sandbox VMs could reach the Proxmox host's `pveproxy` (API, port 8006), SSH, and other host services from inside the VM. This made the Proxmox API a target for cyber-eval agents — they'd spend significant solver budget trying to attack it.

## How VMs reach the host today

The exposure is on at least three IPs from inside any sandbox VM (verified empirically against a live Proxmox 9 host on AWS):

- the SDN gateway IP — `192.168.X.1:8006`, where `X` is the random third octet picked by `SdnCommands.generate_sdn_config`
- any other IP the host owns — e.g. on the EC2 nested-virtualisation path, `vmbr0` is `10.10.10.1/24`, so VMs can also hit `10.10.10.1:8006`
- the host's external NIC IP, reached via the FORWARD rule the SDN already sets up with `snat=True`

A `TODO firewall to block access to proxmox?` was already left in `sdn_commands.py:325` flagging the same gap — this PR removes that TODO.

## The fix

In my opinion this is the responsibility of the component that *builds* the proxmox server, rather than a run-time consideration of the provider. Hence, both scripts (src/proxmoxsandbox/scripts/ec2/launch.sh and src/proxmoxsandbox/scripts/virtualized_proxmox/build_proxmox_auto.sh) set the proxmox firewall to block connections.

Existing AMIs / qcow images will need to be updated.


sinman-aisi:
This has too much arcane knowledge for my little SWE brain to be able to say much useful. 

Do you think it's likely to break anything? We could push a PR with this commit pinned in cyber register and trigger the CI job that runs TLO if you think that would be useful

art-dsit:
> This has too much arcane knowledge for my little SWE brain to be able to say much useful.
> 
> Do you think it's likely to break anything? We could push a PR with this commit pinned in cyber register and trigger the CI job that runs TLO if you think that would be useful

Yes I think this needs a bunch of testing on all the ranges. I haven't reviewed the code very thoroughly yet either, but if it's easy to pin a commit and test, please do

pgiav:
If you don't have a rule allowing TCP/8006 access and turn on the datacenter firewall you will lock yourself out of the PVE UI (and API too, I would assume). Just flagging this because I've fallen in that trap before myself. 

The other rules look fine, on our ranges I always configure these three (all `in | ACCEPT`) at the datacenter level:
1. TCP/8006 
2. Macro -> DNS
3. Macro -> DHCPfwd

Blocking TCP/8006 seems to be the whole intention of this PR but the claudeslop summary doesn't make it clear to me how you'll still allow access from the Inspect process and non-range VMs to PVE.

art-dsit:
OK I've completely changed how the PR works, so that the firewall is enabled at build time rather than runtime

sinman-aisi (approved):

PR #73, opened 2026-06-01
Fix tests broken by inspect_ai update

art-dsit:
## Summary

Two tests broke after bumping `inspect_ai`:

- **`test_inspect_eval`** — `bash()` renamed its tool parameter from `cmd` to `command`. The mock tool-call in `test_eval.py` is updated to match.
- **`test_self_check[linux]` / `test_exec_input_large`** — a new self_check pipes 1 MiB of stdin to `sandbox.exec()`. Our `exec()` inlines stdin (base64-encoded) into the shell script and writes the whole script in a single `agent/file-write` call. With 1 MiB stdin, that POST is ~1.4 MiB and pveproxy closes the connection without a response, surfacing as `httpx.ReadError('')`.

  For stdin above 30 KiB, write it to a separate file via the existing chunked `self.write_file` and redirect from it in the script. Small stdin keeps the inline base64 path unchanged.

## Why 30 KiB and not 512 KiB

The binding limit on inlined stdin isn't pveproxy's POST cap — it's the PVE `agent/file-write` validation on the `content` field. Measured against a Proxmox 8.x instance:

| raw stdin | result |
| --- | --- |
| 32 KiB | OK |
| 45 KiB – 256 KiB | HTTP 400 `value may only be 61440 characters long` |
| 384 KiB+ | `httpx.ReadError('')` (pveproxy POST cap, ~512 KiB) |

The 60 KiB `content` cap caps raw stdin at ~34 KiB after two base64 layers + script overhead. 30 KiB leaves a little headroom for env vars / cwd / user. The 1 MiB self_check case overshoots by enough to skip the API validation and hit pveproxy directly, which is why we saw `ReadError` rather than a 400. Refs: [forum 166200](https://forum.proxmox.com/threads/166200), [forum 105556](https://forum.proxmox.com/threads/105556).

## Test plan

- [x] `tests/proxmoxsandboxtest/test_eval.py::test_inspect_eval` passes
- [x] `tests/proxmoxsandboxtest/test_proxmox_sandbox_agent_commands.py::test_self_check[linux]` passes (includes the new `test_exec_input_large`)
- [x] `ruff check` / `ruff format --check` clean
- [ ] CI


sinman-aisi (approved):
LGTM!

PR #74, opened 2026-06-07
Retry QGA exec/read-back on transport errors & 596/597 broken pipe

sinman-aisi:
## What

The guest-agent retry wrapper (`AgentCommands._retry_on_qga_error`) only caught `httpx.HTTPStatusError` with status **500**. Two common transient failures therefore propagated and killed the entire eval sample (full restart) with no retry:

- **httpx transport errors** — `ConnectError`, `ConnectTimeout`, `ReadTimeout`, `ReadError`, `RemoteProtocolError` (Proxmox API host briefly unreachable / dropped the connection).
- **Proxmox `596`/`597` "Broken pipe"** — returned when streaming a command's `stdout`/`stderr` back via `agent/file-read` (these are 5xx but `!= 500`, so fell straight through).

## Change

- `_is_transient_qga_error` now treats `httpx.TransportError` and any `>= 500` QGA response as transient, **except** a 500 reporting a missing file (it will never appear; `read_file_or_blank` already converts it to empty content).
- `_retry_on_qga_error` catches both `HTTPStatusError` and `TransportError` and retries with bounded exponential backoff (5 attempts, 2s→20s).
- Adds `tests/proxmoxsandboxtest/test_qga_retry.py` — pure unit tests (no infra) covering the retry / no-retry matrix.

## Why

In a batch of long-running Transatomic eval runs, these two classes accounted for ~83% of non-provisioning sample failures; each aborts a multi-hour sample. Companion analysis: AISI CASTCE-4643 / CASTCE-4555. Builds on the intent of #72 and the issue raised in CASTCE-4473.

## Test

`uv run pytest tests/proxmoxsandboxtest/test_qga_retry.py` → 17 passed. Existing e2e tests unchanged (require Proxmox infra).


art-dsit:
## End-to-end testing notes

This testing was done by Claude (via Claude Code). Recording what was covered so reviewers have it.

**Setup.** A real Proxmox host on an AWS bare-metal instance, with nested Proxmox instances vended on it, driven through `inspect eval`. Faults were injected against the nested Proxmox while a sample was mid-run, and each result compared A/B against `main`.

**Transport errors (the original fix).** Making the nested Proxmox API briefly unreachable mid-sample (freezing it, and separately dropping its API port) produces `httpx` transport errors. On `main` these are uncaught and abort the sample; on this branch they are retried and the sample recovers.

**596/597 broken pipe on large reads.** Streaming multi-MB command output back while RST-ing the Proxmox `pveproxy`↔`pvedaemon` connection produces genuine `596` "broken pipe" responses. On `main` these fall straight through (`596 != 500`, no retry) and fail the read; on this branch the read is retried and recovers with the payload byte-intact.

**exec-start idempotency.** Because `agent/exec` is now retried, a retry could in principle launch a command twice (if the guest agent received the request but the pid response was lost). The wrapper script guards the command with `flock`, so any number of duplicate launches run the command exactly once and each only returns once the result is ready. Verified two ways: running the generated wrapper script concurrently, and driving two concurrent `agent/exec` calls for one script through a real guest agent — in both cases the command body ran once and the captured stdout/return code were intact.

**Regression.** The full end-to-end test suite passes against a vended Proxmox (122 passed, 4 skipped).

art-dsit:
### Fault-injection methods (for reproducibility)

For anyone wanting to reproduce the above. The nested Proxmox is a libvirt domain on the bare-metal host, and its QEMU does a host port-forward (`metal:PORT -> guest:8006`) for the API. All commands below run on the bare-metal host unless noted.

**Freeze / unfreeze the nested Proxmox** (API becomes unreachable; in-flight calls time out, new connects fail):
```bash
virsh suspend <domain>
sleep <secs>
virsh resume <domain>
```

**Throttle its CPU** (slow API, exercises read timeouts / graceful degradation):
```bash
virsh schedinfo <domain> --live --set vcpu_quota=1000 --set vcpu_period=100000   # ~1%
# ... restore ...
virsh schedinfo <domain> --live --set vcpu_quota=-1
```

**Drop the API port** (realistic "host briefly unreachable" blip; guest keeps running, only the API path is down). A `trap` guarantees the rule is removed even if the script is killed:
```bash
trap 'iptables -D INPUT -p tcp --dport <PORT> -j DROP 2>/dev/null' EXIT
iptables -I INPUT -p tcp --dport <PORT> -j DROP
sleep <secs>
```

**Induce a genuine `596`/`597` broken pipe.** These come from `pveproxy` losing its connection to `pvedaemon` (TCP on `127.0.0.1:85`) while streaming a response, so the RST has to happen *inside* the nested Proxmox. I opened a temporary shell into it by adding an SSH forward through the QEMU monitor, then RST-ed the active `:85` connection mid-read (one RST is enough to produce one `596`):
```bash
# on the bare-metal host: expose the guest's sshd
virsh qemu-monitor-command <domain> --hmp 'hostfwd_add tcp::2222-:22'
# then, inside the nested Proxmox (ssh -p 2222), while a large read streams:
ss -K dst 127.0.0.1 dport = 85       # RSTs the pveproxy<->pvedaemon socket
```
(Note: `ss -K` rejected an `( sport = 85 or dport = 85 )` filter; the `dst ... dport = 85` form works.)

**exec-start idempotency** was checked without needing the rare lost-response timing, by reproducing the *condition* it causes (two concurrent launches of one script): running the generated wrapper twice concurrently with `asyncio.gather`, both locally on the emitted `sh` and through two concurrent `agent/exec` calls on a real VM, then asserting an observable side effect fired once and the captured output/return code were intact.

These were thin throwaway scripts (a `virsh`/`iptables`/`ss` wrapper and a couple of `inspect`/QGA harnesses); happy to clean any up into the repo if useful, but they're not part of the diff.

PR #78, opened 2026-06-09
fix: large tool output no longer crashes on truncation

### LINKED ISSUE #77: large tool output can crash
Symptom:

`invalid escape at line 1 column 10493953`

Should be covered by unit tests already but evidently not

art-dsit:
fix: large tool output no longer crashes on truncation

Fixes #77.

## Problem

Large `exec` output crashed with `invalid escape at line 1 column 10493953` (and, for other output, `invalid unicode code point` at the same column).

Output over `MAX_EXEC_OUTPUT_SIZE` (10 MiB) is read back by streaming the QGA `agent/file-read` response in `AsyncProxmoxAPI.read_file`. When the stream exceeded the limit, the old code truncated by hand-appending a `"` and parsing with `from_json(..., allow_partial=True)`:

```python
truncated_json = from_json(b"".join(chunks) + b'"', allow_partial=True)
```

The cut (rounded up to an 8192 chunk boundary = 10493952 bytes) frequently lands **inside a multi-byte sequence** on the wire:
- control bytes JSON-escape to 6-byte `\uXXXX` escapes;
- high bytes come back as 2-byte sequences — Proxmox base64-decodes the guest-agent payload into a Perl byte string and its JSON encoder re-encodes each byte `0x80-0xFF` via latin-1 (the read-side twin of the `Wide character` mangling the `write_file` path already documents).

Closing the string with `"` then produces invalid JSON (`\u00"`, or a lone UTF-8 lead byte followed by `"`), so `from_json` raises. (file-read responses are also gzipped, so the early `content-length > max_size` guard compares the *compressed* size and rarely fires — the streaming/truncation path is the normal path for large output.)

## Fix

Parse the partial body with `allow_partial="trailing-strings"` and drop the manual `"`, so pydantic_core discards the incomplete trailing token cleanly. Extracted into `_parse_truncated_file_read`.

Since #80 landed (un-mangling full `exec` stdout/stderr), this PR also decodes the **truncated** output consistently: a new `_read_exec_output` helper catches `OutputLimitExceededError` and runs its `truncated_output` through the same `_recover_qga_bytes` + UTF-8 decode as a full read. Otherwise `OutputLimitExceededError.truncated_output` would have leaked Proxmox's mangled wire form (mojibake) even though a non-truncated read would not.

## Why the existing test missed it

`test_exec_10mb_limit` emitted 10 MiB of `'a'` and stayed *under* the limit — it never truncated, and plain ASCII has no multi-byte sequence for the cut to split.

## Test

Reworked `test_exec_10mb_limit` to push output past the limit with a **4-byte UTF-8 character** (😀, U+1F600) and parametrized it over two outputs differing by a single leading byte. Every byte of the emoji is non-ASCII, so the whole payload is uniform 2-byte wire pairs; the single lead byte flips parity, so exactly one variant's cut lands mid-sequence and reproduces the crash. The assertion checks the recovered, **decoded** content is real emoji text (tolerating at most one trailing `U+FFFD` where the size cut falls inside a 4-byte sequence), not the mangled byte form.

Verified end-to-end against a real Proxmox instance — both `even-offset` and `odd-offset` variants pass (`2 passed`).

## Also in this PR

A latent cloud-init check in `built_in_vm.py` failed during the e2e run on a slower instance: `cloud-init status --wait` prints progress dots before the final status line, and the check compared the whole stripped output to `"status: done"`. Changed to `.endswith("status: done")` so the dots no longer cause a false `cloud-init failed` (and `status: error`/`degraded` still fall through to the raise). Untouched since the first commit; no code change caused it, instance timing exposed it.


art-dsit:
The proxmox-side mojibake fix flagged here ("a proxmox-side fix to follow") is now up as a draft: #80.

It only decodes successfully-read exec stdout/stderr, so it shouldn't touch this PR's truncation / `OutputLimitExceededError` preview path or its `Ã` assertion — but worth a check on whichever of the two merges second.

sinman-aisi (approved):
Nice, thanks!

PR #79, opened 2026-06-09
fix: surface exec timeout instead of opaque RetryError (#76)

### LINKED ISSUE #76: QGA contends with exec for resources
It is sometimes observed that QGA dies/fails to respond if the agent is doing something intensive like `john`. The stack trace is uninformative in this case

```│ /home/ubuntu/cyber-register/.venv/lib/python3.14/site-packages/proxmoxsandbox/_proxmox_sandbox_environment.py:700 in │
│ exec                                                                                                                 │
│                                                                                                                      │
│   697 │   │   │   self.TRACE_NAME,                                                                                   │
│   698 │   │   │   f"exec_command {self.vm_id=} {exec_response_pid=}",                                                │
│   699 │   │   ):                                                                                                     │
│ > 700 │   │   │   exec_status = await wait_for_exec(self.vm_id, exec_response_pid)      
```

with the error message `RetryError: RetryError[<Future at 0x70682662f750 state=finished returned bool>]`

We should confirm if this is the case, and definitely improve the error message

art-dsit:
## Issue

Fixes #76 — `exec` sometimes fails with an opaque `tenacity.RetryError: RetryError[<Future ... returned bool>]` and an uninformative stack trace.

## What I observed

I built a repro that runs through `inspect eval` against a Proxmox VM and ran it on the current code before changing anything. In that setup:

- The error reproduces for a command that **handles SIGTERM and does not exit immediately** (e.g. `trap 'sleep 30' TERM; while :; do :; done`, a stand-in for `john` saving its session) when it runs past its timeout.
- A command that exits promptly on SIGTERM times out cleanly (`SandboxTimeoutError`) on the same code.
- I did **not** observe the guest agent becoming unresponsive under CPU saturation (8 busy loops on a 2-vCPU VM): cheap probe `exec`s kept returning in ~2s throughout. So, at least in my testing, the trigger appears to be the command's SIGTERM handling rather than QGA itself failing — though I can't rule out that heavy real workloads behave differently from my synthetic ones.

Mechanism, as far as I can tell: `exec` wraps the command in-guest with `timeout -k 5s {timeout}s`, which SIGTERMs at `timeout` and only SIGKILLs at `timeout`+5s. The outer poll loop used `stop_after_delay(timeout)` — the same value — so once the deadline passes, the next poll can catch the command still alive in the SIGKILL grace window and raise `RetryError` instead of seeing the real exit.

## Change

- Extend the poll deadline by a grace margin (`timeout + 8s`) so the poll outlasts the in-guest SIGKILL and observes the command's actual exit.
- Convert any residual `RetryError` into a `TimeoutError` carrying VM/pid/timeout/command context, so a genuinely stuck poll no longer surfaces as an opaque tenacity error.

## Caveat

A command that ignores SIGTERM past the grace period is SIGKILLed and returns `rc=137` as a normal `ExecResult` rather than raising `TimeoutError`. The common case (a command that exits within the grace, including `john` as far as I understand it) returns `rc 124` → `TimeoutError`. I did not map `137` to `TimeoutError` because that signal is ambiguous (OOM, external kill); happy to revisit if a timeout-shaped error is preferred there.

## Testing

- e2e regression test (`test_exec_timeout_with_sigterm_handler_no_retryerror`): fails on the pre-change code with `RetryError`, passes with the change. Runs against a real Proxmox VM.
- Unit test (`test_exec_timeout.py`) for the `RetryError`→`TimeoutError` backstop using a mocked unresponsive agent.
- Existing `test_exec_10mb_limit` still passes e2e.
- These were run against a single Linux VM on one NitroVirt instance; I have not exercised the Windows path or multi-instance pools.

### DIFF
--- a/src/proxmoxsandbox/_proxmox_sandbox_environment.py
+++ b/src/proxmoxsandbox/_proxmox_sandbox_environment.py
@@ -43,13 +43,8 @@ from proxmoxsandbox.schema import (
 # for derivation); 30 KiB leaves a little headroom for env/cwd/etc. overhead.
 _INLINE_STDIN_LIMIT = 30 * 1024
 
-# Grace added to the exec polling deadline on top of the caller's timeout.
-# The in-guest `timeout -k 5s {timeout}s` wrapper only SIGTERMs the command at
-# `timeout`; a command that doesn't exit at once on SIGTERM (e.g. `john` saves
-# its session first) lingers until the SIGKILL at `timeout`+5s. Polling for only
-# `timeout` then catches the command mid-shutdown and raised an opaque
-# RetryError (issue #76); the grace lets the poll outlast the SIGKILL so the
-# real exit (rc 124, or 137 if it had to be killed) is seen instead.
+# Poll past the in-guest `timeout -k 5s` SIGKILL (at timeout+5s) so we observe
+# the real exit instead of catching the command mid-shutdown.
 _EXEC_POLL_GRACE_SECONDS = 10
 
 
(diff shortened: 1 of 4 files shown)


src/proxmoxsandbox/_proxmox_sandbox_environment.py
```diff
@@ -43,6 +43,10 @@
 # for derivation); 30 KiB leaves a little headroom for env/cwd/etc. overhead.
 _INLINE_STDIN_LIMIT = 30 * 1024
 
+# Poll past the in-guest `timeout -k 5s` SIGKILL (at timeout+5s) so we observe
+# the real exit instead of catching the command mid-shutdown.
+_EXEC_POLL_GRACE_SECONDS = 10
```
dpolatajko-aisi: nit: i suppose this is coupled to the 5s timeout - might be worth expressing one as a multiple of the other in case it changes

dpolatajko-aisi (approved):
reproduced and verified - barring one nitpick i think this is good

src/proxmoxsandbox/_proxmox_sandbox_environment.py
art-dsit: yeah I actually didn't particularly like the 2x multiple, wondering if e.g. 8s (5 + 3) would be better

art-dsit:
Will merge this once my test run finishes

PR #80, opened 2026-06-15
Decode UTF-8 in exec stdout/stderr (proxmox-side mojibake fix)

art-dsit:
## Problem

`exec` stdout/stderr came back as mojibake for non-ASCII output — `café` arrived as `cafÃ©`.

Proxmox's `agent/file-read` returns the file's bytes with each raw byte mapped to a Latin-1 codepoint, then serialised as UTF-8 JSON (the read-side twin of the `Wide character` mangling the `write_file` path already documents — it's the JSON encoder's `utf8 => 1` applied to a flagless Perl byte string). `read_file` already reverses this (`encode("iso-8859-1")`), but `exec()` read `script.stdout`/`script.stderr` straight from the QGA `content` and used it raw, so the recovery never happened.

## Fix

- Factor the recovery into `_recover_qga_bytes` and share it between `read_file` and a new `_decode_exec_output`.
- `_decode_exec_output` decodes UTF-8 with `errors="replace"`, substitute rather than crash (The contract does allow us to throw UnicodeDecodeError here, but given this has been returning garbage for goodness knows how long, I'm happy with some replacement characters in the output)


art-dsit:
The out-of-scope limit/`decode=0` follow-up is now tracked in #81.

src/proxmoxsandbox/_proxmox_sandbox_environment.py
```diff
@@ -44,6 +44,13 @@
 _INLINE_STDIN_LIMIT = 30 * 1024
 
 
+def _recover_qga_bytes(mangled: str) -> bytes:
+    # Proxmox's file-read returns each raw file byte as a Latin-1 codepoint, then
+    # serialises that as UTF-8 JSON, so non-ASCII bytes arrive double-encoded
+    # (e.g. "é" -> "Ã©"). Encoding back to ISO-8859-1 recovers the original bytes.
```
art-dsit, line 50: This is true and strange

src/proxmoxsandbox/_impl/agent_commands.py
```diff
@@ -148,7 +148,13 @@ async def write_file(self, vm_id: int, content: bytes, filepath: str):
             )
 
     async def exec_command(self, vm_id: int, command: List[str]):
-        """Execute a command in the VM using QEMU agent."""
+        """Execute a command in the VM using QEMU agent.
+
+        Every element of `command` MUST be ASCII. A single non-ASCII byte in
+        the QMP guest-exec arguments breaks Proxmox's agent bridge: the call
+        times out and the next one reports "QEMU guest agent is not running"
+        (upstream Bugzilla #6609, still open).
```
bmillwood-aisi:
```suggestion
        (upstream bug report: https://bugzilla.proxmox.com/show_bug.cgi?id=6609 ).
```
"still open" might hopefully go out of date one day

src/proxmoxsandbox/_proxmox_sandbox_environment.py
bmillwood-aisi: perhaps worth writing a test?

art-dsit: It's now covered by inspect core's self_check https://github.com/UKGovernmentBEIS/inspect_ai/pull/4225

sinman-aisi:
Feels like we could probably test this? Don't know what the best way would be. Is it possible to test it without a real proxmox instance?

art-dsit:
> Feels like we could probably test this? Don't know what the best way would be. Is it possible to test it without a real proxmox instance?

Tested by https://github.com/UKGovernmentBEIS/inspect_ai/pull/4225 when you run this repo's test suite

sinman-aisi (approved):

PR #82, opened 2026-06-16
EC2: derive Proxmox node name from the cloud-init hostname

art-dsit:
## What
The EC2 AMI's `proxmox-ami-fixup-hostname` force-set the hostname to `proxmox` every boot, so the PVE node was always named `proxmox`. This changemakes the node name **follow the OS hostname**, so a launcher that gives the box a real hostname gets a node identifiable by that name (in the UI and in metrics).

Deliberately makes **no assumption** that cloud-init sets a friendly name:
- **If** the launcher set a hostname (e.g. a cloud-config `hostname: <name>` with `preserve_hostname: false`), the node takes that name.
- **Otherwise** (plain EC2 launch) the hostname is the private-DNS default `ip-x-x-x`; the fixup detects that and keeps `proxmox`. Default/upstream behaviour is unchanged.

The node is no longer always `proxmox`; set `PROXMOX_NODE` to the hostname you give the box, or `/nodes/proxmox/...` calls 404.

### DIFF
--- a/src/proxmoxsandbox/scripts/ec2/userdata.sh
+++ b/src/proxmoxsandbox/scripts/ec2/userdata.sh
@@ -239,26 +239,36 @@ pvesm set local --content images,rootdir,vztmpl,backup,iso,snippets,import
 # --- AMI boot-time fixup services ---
 # When an AMI is launched with a new IP, EC2 changes the hostname to ip-x-x-x-x,
 # breaking Proxmox node identity, SSL certs, and pveproxy. These services fix
-# that on every boot. The password fixup additionally detects fresh launches
-# (by EC2 instance-id) and regenerates the root password so credentials don't
-# leak across instances launched from the same AMI.
-NODE_NAME="proxmox"
+# that on every boot: the hostname fixup pins the node name to the cloud-init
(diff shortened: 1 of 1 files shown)


art-dsit:
This doesn't work if you've baked an AMI with template VMs in; there's extra fiddling needed, per

https://pve.proxmox.com/wiki/Renaming_a_PVE_node

https://www.derekfarr.com/docs/articles/rename-proxmox-node/

https://www.threatbear.co/blog/properly-renaming-proxmox-nodes/

art-dsit:
(claude comment)

Closing in favour of #83.

The node rename was meant to give the CloudWatch metrics from #83 a per-box identity. But renaming the PVE node strands any preloaded VM templates under `/etc/pve/nodes/<old-name>/` — pmxcfs keys VM configs by node name, and AISI prepares AMIs with templates baked under node `proxmox`.

#83 gets the same per-box identity from the OTel collector's `resourcedetection` (EC2 instance-id + `ec2.tag.Name`) with the node left as `proxmox`, so templates survive. Verified end-to-end against the CloudWatch PromQL store (`@resource.ec2.tag.Name=art-otel-test`, `node=proxmox`).

PR #84, opened 2026-06-17
fix: handle `exec` killing it's own wrapper process

### LINKED ISSUE #75: exec can kill its own wrapper process
It seems possible that an exec command that greps and pkills might sometimes find and kill its own wrapper script, confusing the provider and causing an exception

dpolatajko-aisi:
Closes #75 

## Description

This PR explicitly handles the case where the command on the sandbox kills it's own wrapper process, by checking exit codes when process before the return code was written, and returning the relevant exit code rather than allowing errors to escalate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Fix: `exec()` no longer aborts the sample when a command kills its own command-runner wrapper process (e.g. `pkill -f`); it returns a failed `ExecResult` (`128+signal`, or `137` when the signal is unavailable) instead of raising a misleading `TimeoutError` (#75)
+- Fix: `exec()` no longer aborts the sample when a command kills its own command-runner wrapper process (e.g. `pkill -f`); it returns a failed `ExecResult` (`128+signal`, or `137` when the signal is unavailable) instead of raising a misleading `TimeoutError`
 
 ## 0.11.0 - 2026-06-01
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Fix: `exec()` no longer aborts the sample when a command kills its own command-runner wrapper process (e.g. `pkill -f`); it returns a failed `ExecResult` (`128+signal`, or `137` when the signal is unavailable) instead of raising a misleading `TimeoutError` (#75)
+
 ## 0.11.0 - 2026-06-01
 
 - Faster large `write_file` on Linux via hot-plugged ISO (reserves `sata5` on sandbox VMs); falls back to the guest-agent path on failure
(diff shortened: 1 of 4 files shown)


art-dsit:
(From Claude) Suggestion on `_read_return_code`: rather than returning `int | None`, raise a named exception for the "wrapper never wrote a returncode" case and use `reraise=True`.

```python
class ReturnCodeNotWritten(Exception):
    """Wrapper exited without writing its returncode file."""

@tenacity.retry(
    wait=tenacity.wait_exponential(min=0.1, exp_base=1.3),
    stop=tenacity.stop_after_delay(2),
    reraise=True,
)
async def _read_return_code(self, tmp_start) -> int:
    content = (await self.agent_commands.read_file_or_blank(...))["content"].strip()
    if not content:
        raise ReturnCodeNotWritten()
    return int(content)
```

The caller catches `ReturnCodeNotWritten` for the killed branch.

Two reasons:

- The return type is plain `int`; the missing-code case is named rather than encoded as a falsy value.
- `reraise=True` reraises the last actual exception, so a persistent read/transport failure surfaces as itself instead of being folded into the same `None` and mislabelled "killed by the command itself".

Timeouts are unaffected: a genuine timeout writes `124` into the returncode file (via the `timeout` prefix), so `_read_return_code` reads it normally and returns `124`. Only a killed wrapper leaves the file missing/empty, which is the one case that should hit the killed branch.

dpolatajko-aisi:
> (From Claude) Suggestion on `_read_return_code`: rather than returning `int | None`, raise a named exception for the "wrapper never wrote a returncode" case and use `reraise=True`.
> 
> ```python
> class ReturnCodeNotWritten(Exception):
>     """Wrapper exited without writing its returncode file."""
> 
> @tenacity.retry(
>     wait=tenacity.wait_exponential(min=0.1, exp_base=1.3),
>     stop=tenacity.stop_after_delay(2),
>     reraise=True,
> )
> async def _read_return_code(self, tmp_start) -> int:
>     content = (await self.agent_commands.read_file_or_blank(...))["content"].strip()
>     if not content:
>         raise ReturnCodeNotWritten()
>     return int(content)
> ```
> 
> The caller catches `ReturnCodeNotWritten` for the killed branch.
> 
> Two reasons:
> 
> * The return type is plain `int`; the missing-code case is named rather than encoded as a falsy value.
> * `reraise=True` reraises the last actual exception, so a persistent read/transport failure surfaces as itself instead of being folded into the same `None` and mislabelled "killed by the command itself".
> 
> Timeouts are unaffected: a genuine timeout writes `124` into the returncode file (via the `timeout` prefix), so `_read_return_code` reads it normally and returns `124`. Only a killed wrapper leaves the file missing/empty, which is the one case that should hit the killed branch.

yeah i agree that this is a better pattern, nice

art-dsit (approved):

PR #85, opened 2026-06-19
EC2: trim CloudWatch OTLP metrics to memory/CPU/disk

art-dsit:
pvestatd emits ~1400 datapoints/cycle; the bulk is per-VM-per-disk proxmox_vm_blockstat_*, which exceeds CloudWatch's 1000-datapoints-per-request limit (the OTLP exporter returns HTTP 400 and drops the whole batch).

Add a filter processor to the embedded collector pipeline that keeps only memory, CPU and disk for guests and host: capacity (vm disk/maxdisk, node rootfs, storage), plus per-device I/O (rd/wr operations + total_time_ns for IOPS and latency) and io PSI (pressureio*) for spotting I/O-induced stalls. Also hard-cap each export request at 1000 datapoints via the batch processor as a backstop if the guest count grows.


sinman-aisi (approved):
Thank you!

PR #87, opened 2026-06-19
fix: chunk file writes on exec

### LINKED ISSUE #86: exec(): large commands trip the 61440-char QGA file-write cap because the wrapper script isn't chunked
_Analysis and description by Claude (Claude Code)._

## Summary
`exec()` serialises the command into a wrapper script (`script.sh` / `script.bat`) and uploads it with `_write_file_only`, which issues a single `agent/file-write`. That endpoint caps the base64 `content` at 61440 chars, so a command whose script exceeds ~45 KiB raw fails with `HTTP 400 Parameter verification failed: value may only be 61440 characters long`, killing the call. A model that inlines a large blob into one command (a big base64 heredoc, `python -c "<lots of code>"`, etc.) hits this.

## Why
- `exec()` writes the wrapper via `_write_file_only(...)` → `AgentCommands.write_file` → one POST, no chunking.
- The public `SandboxEnvironment.write_file()` already chunks (plus the ISO fast path), and large `stdin` is already routed to a file above `_INLINE_STDIN_LIMIT` (30 KiB) precisely to dodge this cap — but the command embedded in the wrapper script has no equivalent guard. It's the one remaining un-chunked write on the `exec` path.

## Repro
```python
await sandbox.exec(["echo", "-n", "a" * (100 * 1024)])
```
→ HTTP 400 from the file-write of `script.sh` (the ~133 KiB base64 wrapper exceeds 61440). Same shape on Windows for `script.bat`.

## Related
- #81, #77 — the read side of the same QGA latin-1/base64 wire-encoding limits.

## Fix direction
Upload the wrapper script with only the bootstrap-safe primitives (one `agent/file-write` per chunk + one `agent/exec`), never via the public `write_file`/`exec` (which would re-enter the exec primitive and recurse). For oversize scripts, write `≤40 KiB` parts and reassemble them inside the single launched command (`cat …part* > script && exec sh script`), so the existing exec-status poll covers the combine with no extra round-trip. Windows uses `copy /b` + `cmd.exe /c`.

dpolatajko-aisi:
Closes #86 

## Description

Chunks the file write for large scripts, to avoid hitting a character limit when writing to the server. Reassembles the chunks as part of the execution command that is issued to the server.

### DIFF
--- a/src/proxmoxsandbox/_proxmox_sandbox_environment.py
+++ b/src/proxmoxsandbox/_proxmox_sandbox_environment.py
@@ -43,6 +43,10 @@ from proxmoxsandbox.schema import (
 # for derivation); 30 KiB leaves a little headroom for env/cwd/etc. overhead.
 _INLINE_STDIN_LIMIT = 30 * 1024
 
+# Raw bytes per exec-wrapper-script chunk; base64 stays under the 61440-char
+# agent/file-write cap.
+_SCRIPT_CHUNK_SIZE = 40 * 1024
+
 _IN_GUEST_KILL_GRACE = 5
 
 # Poll past the in-guest `timeout -k {_IN_GUEST_KILL_GRACE}s` SIGKILL (at timeout+5s)
@@ -714,9 +718,11 @@ class ProxmoxSandboxEnvironment(SandboxEnvironment):
                 timeout=timeout,
             )
             script_path = f"{tmp_start}script.bat"
-            await self._write_file_only(script_path, script)
+            launch = await self._upload_exec_script(
+                script_path, script, is_windows=True
+            )
             exec_post_response = await self.agent_commands.exec_command(
-                vm_id=self.vm_id, command=["cmd.exe", "/c", script_path]
+                vm_id=self.vm_id, command=launch
             )
         else:
             # Inlined stdin is base64-encoded into the script, which itself
@@ -753,9 +759,11 @@ class ProxmoxSandboxEnvironment(SandboxEnvironment):
                 timeout=timeout,
                 stdin_file=stdin_file,
             )
-            await self._write_file_only(f"{tmp_start}script.sh", script)
+            launch = await self._upload_exec_script(
+                f"{tmp_start}script.sh", script, is_windows=False
+            )
             exec_post_response = await self.agent_commands.exec_command(
-                vm_id=self.vm_id, command=["sh", f"{tmp_start}script.sh"]
+                vm_id=self.vm_id, command=launch
             )
 
         exec_response_pid = exec_post_response["pid"]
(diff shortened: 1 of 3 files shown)


art-dsit:
Good find. These agents are crazy, eh?

Rather than unit-testing here can we please add to https://github.com/UKGovernmentBEIS/inspect_ai/blob/main/src/inspect_ai/util/_sandbox/self_check.py ? Then all sandboxes will get the same test coverage.

Looks like the max command line size on linux is about 2MB  (old discussion [here](https://serverfault.com/questions/163371/linux-command-line-character-limit); for me `getconf ARG_MAX`
yields 2097152) so maybe 1MB would an appropriate size?

src/proxmoxsandbox/_proxmox_sandbox_environment.py
```diff
@@ -931,6 +939,49 @@ async def _write_file_only(self, file: str, contents: str | bytes) -> None:
...
+                else ["sh", script_path]
+            )
+
+        chunks = [
+            data[i : i + _SCRIPT_CHUNK_SIZE]
+            for i in range(0, len(data), _SCRIPT_CHUNK_SIZE)
+        ]
+        width = len(str(len(chunks) - 1))
+        for i, chunk in enumerate(chunks):
+            await self._write_file_only(f"{script_path}.part{i:0{width}d}", chunk)
```
art-dsit, line 979: This seems duplicated from the existing write_file chunking, and for really large scripts we want to use the CD attachment method. Is there a reason (recursion?) we can't reuse that code?

dpolatajko-aisi: the reason to not reuse it en masse is recursion. i could probably make a reusable abstraction though - which would make sense if there's other logic we'd want to reuse

art-dsit: It would be nice to reuse it as the CD attachment method is much faster and there is some fallback logic if it doesn't work, that we'd want to reuse

dpolatajko-aisi:
> Looks like the max command line size on linux is about 2MB (old discussion [here](https://serverfault.com/questions/163371/linux-command-line-character-limit); for me `getconf ARG_MAX`
> yields 2097152) so maybe 1MB would an appropriate size?

my understanding is that the limit here is proxmox specific, and is the same as the existing chunked write logic (i've now consolidated those constants to make the coupling clearer)

art-dsit:
> > Looks like the max command line size on linux is about 2MB (old discussion [here](https://serverfault.com/questions/163371/linux-command-line-character-limit); for me `getconf ARG_MAX`
> > yields 2097152) so maybe 1MB would an appropriate size?
> 
> my understanding is that the limit here is proxmox specific, and is the same as the existing chunked write logic (i've now consolidated those constants to make the coupling clearer)

Ah sorry should have been clearer - I meant 1MB would be an appropriate size for the self_check test; if we used, say, 5Mb, that would presumably blow up even in the absence of a sandbox transport due to being over Linux's CLI character limit.

dpolatajko-aisi:
> > > Looks like the max command line size on linux is about 2MB (old discussion [here](https://serverfault.com/questions/163371/linux-command-line-character-limit); for me `getconf ARG_MAX`
> > > yields 2097152) so maybe 1MB would an appropriate size?
> > 
> > 
> > my understanding is that the limit here is proxmox specific, and is the same as the existing chunked write logic (i've now consolidated those constants to make the coupling clearer)
> 
> Ah sorry should have been clearer - I meant 1MB would be an appropriate size for the self_check test; if we used, say, 5Mb, that would presumably blow up even in the absence of a sandbox transport due to being over Linux's CLI character limit.

i see - see: https://github.com/UKGovernmentBEIS/inspect_ai/pull/4312

art-dsit (approved):

PR #88, opened 2026-06-26
Block Proxmox sandbox VMs from accessing EC2 instance metadata

ericwinsor-aisi:
## Summary

- Require IMDSv2 with a one-hop response limit for EC2 Proxmox hosts.
- Install persistent IPv4 and IPv6 forwarding blocks for cloud metadata in
  both bundled provisioning paths.
- Extend host-isolation tests and document the migration requirement for
  existing AMIs and templates.

## Security rationale

Sandbox VMs receive outbound access through SNAT. Without an explicit deny,
they can reach EC2 IMDS and potentially retrieve credentials belonging to the
host.

The forwarding rules drop guest traffic before routing or SNAT. Host-originated
traffic does not traverse PREROUTING, so Proxmox services retain IMDS access.

## Deployment note

Existing Proxmox AMIs and virtualized templates must be rebuilt, or the
documented persistent firewall rules must be applied manually.

## Verification

- `uv run pytest tests/proxmoxsandboxtest/test_provisioning_security.py -q`
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run mypy`
- Shell syntax checks for both provisioners and the EC2 launcher
- Live EC2/Proxmox test with effective metadata hop limit 2:
  - host IMDS access remained available
  - sandbox IMDS access was blocked
  - rules survived Proxmox firewall restarts
  - end-to-end isolation test passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Fix: `exec()` no longer aborts the sample when a command kills its own command-runner wrapper process (e.g. `pkill -f`); it returns a failed `ExecResult` (`128+signal`, or `137` when the signal is unavailable) instead of raising a misleading `TimeoutError`
+- Prevent sandbox VMs from accessing EC2 instance metadata credentials when using the bundled provisioning scripts. Rebuild existing Proxmox AMIs/templates or apply the documented host rules to install this protection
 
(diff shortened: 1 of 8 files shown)


art-dsit:
Thanks very much for this! I've extended it a bit in #92 , would love you to take a look at that please

PR #89, opened 2026-06-27
fix: anchor SDN zone ownership regex used by cleanup

ericwinsor-aisi:
## Discovered issue

`ZONE_REGEX = "...[0-9]{3}z"` is the ownership boundary that
`cleanup_no_id` and the `sample_init` pre-check use to decide which SDN
zones the provider owns and may delete. It is unanchored and applied
with `re.match`, so it accepts any zone whose name *starts* with seven
conforming characters. A pre-existing zone such as `abc123za` is
therefore swept — its subnets, VNets, and the zone itself are deleted —
during ordinary, non-interactive sample startup or partial-failure
cleanup. On a shared or long-lived Proxmox host this crosses the
provider's ownership boundary and can take down unrelated tenants'
networking.

The pattern has been unanchored since the initial open-source release;
the prefix-match behaviour appears accidental (the surrounding comments
already say only provider-named zones are eligible).

## Security-model assumptions

The Proxmox host and API credentials are trusted. Pre-existing
user-managed SDN zones (referenced via `sdn_config=None`) and the
static `inspvm*` SDN are intentional and must persist across samples.
Provider-created ephemeral zones are always exactly 7 characters
(`{task_name[:3] sanitised to [a-z0-9]}{000-999}z`). Name-pattern
ownership remains a heuristic; tag- or ID-based ownership would be a
stronger boundary but is out of scope here.

## Remediation

- Replace `ZONE_REGEX` with the anchored, explicit
  `r"^[a-z0-9]{3}[0-9]{3}z$"`.
- Add `is_ephemeral_zone()` using `re.fullmatch` and route every call
  site through it (`infra_commands.cleanup_no_id`,
  `_proxmox_sandbox_environment._ensure_instance_clean`,
  `sdn_commands.create_sdn` sanity check). Removes the inline
  `re.match` callers so the same mistake cannot be reintroduced at a
  new site.
- The `[a-z0-9]{3}` tightening is safe: zone IDs are built from
  `re.sub("[^a-zA-Z0-9]", "x", task_name[:3].lower())`, so every
  provider-generated name still satisfies the `create_sdn` sanity
  check.
- Add regression tests for near-miss names (`abc123za`, `abc123z-prod`,
  `abc123zz`, `xabc123z`) at the predicate, `cleanup_no_id`, and
  `sample_init` pre-check layers.

## Verification

`ruff check`, `ruff format --check`, `mypy`, and the non-live unit
suites (`test_cleanup_no_id`, `test_sample_init_cleanup`,
`test_cli_cleanup`, plus schema/pool/config/task-wrapper) all pass. The
existing mock-backed reproduction now fails at its first assertion: the
near-miss zone no longer matches.


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Security fix: anchor the ephemeral SDN zone regex so automatic cleanup cannot delete unrelated pre-existing zones whose names share a 7-character prefix with provider zones
```
bmillwood-aisi:
```suggestion
- Anchor the ephemeral SDN zone regex so automatic cleanup is less likely to delete unrelated pre-existing zones whose names share a 7-character prefix with provider zones
```

the idea is that this is an improvement but it's not really a guarantee, so I wouldn't call it necessarily more secure

src/proxmoxsandbox/_impl/sdn_commands.py
```diff
@@ -28,7 +28,23 @@
...
 VnetAliases: TypeAlias = List[Tuple[str, str | None]]
 
-ZONE_REGEX = "...[0-9]{3}z"
+# Ephemeral zones the provider creates are named exactly
+# `{task_name[:3] sanitised to [a-z0-9]}{000-999}z`. This pattern is the
+# ownership boundary for automatic, non-interactive cleanup, so it must be
+# anchored at both ends — a prefix match would let a pre-existing zone such
+# as `abc123za` be swept and deleted as if the provider owned it. Use
+# `is_ephemeral_zone()` rather than matching this directly.
+ZONE_REGEX = r"^[a-z0-9]{3}[0-9]{3}z$"
```
bmillwood-aisi:
```suggestion
ZONE_REGEX = r"^[a-z0-9]{1,3}[0-9]{3}z$"
```

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Anchor the ephemeral SDN zone regex so automatic cleanup is less likely to delete unrelated pre-existing zones whose names share a 7-character prefix with provider zones
```
bmillwood-aisi:
```suggestion
- Anchor the ephemeral SDN zone regex so automatic cleanup is less likely to delete unrelated pre-existing zones
```

bmillwood-aisi (approved):

PR #92, opened 2026-06-30
Generalise sandbox isolation: RFC 3927 link-local block + disable guest IPv6

art-dsit:
## Summary

Supercedes #88: generalises the bundled provisioning scripts beyond EC2.

- **Enforce RFC 3927 in both provisioners** (`build_proxmox_auto.sh` and
  `scripts/ec2/userdata.sh`): a router must not forward IPv4 link-local
  `169.254.0.0/16`, but Linux forwards it anyway by default when routing
- **Disable IPv6 for sandbox guests** IPv6 is unsupported for guests on this provider, so this removes an unmanaged parallel stack 
## Verification

End-to-end against real **metal** (`build_proxmox_auto.sh` → vended) and
**nitrovirt** (`userdata.sh`) hosts:

- Full `pytest tests/` suite passes on both
- Host retains IMDS (`200`) with the shipped rules; moving the source drop to
  `raw PREROUTING` was shown to take host IMDS to `000` — the breakage the
  `FORWARD` placement avoids.
- Sandbox guest blocked from IMDS on both `vmbr0` and SDN vnets; removing the
  rules lets the guest reach IMDS (`401`), confirming the destination drop is
  load-bearing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,17 +52,16 @@ them are isolated out of the box.
 If you provision Proxmox some other way, configure equivalent persistent rules
 **on the node**. The Proxmox rules accept management ports only on the
 default-route interface (where external callers arrive) and leave SDN DNS/DHCP
-open. The remaining rules enforce RFC 3927 section 7, which Linux's forward path
-ignores: a router must not forward IPv4 link-local (`169.254.0.0/16`) traffic, but
-Linux forwards it anyway, so dropping it stops a sandbox guest reaching the host's
-cloud metadata service — and any other link-local endpoint — over its on-link route. The
+open. The remaining rules enforce RFC 3927 section 7: a router must not forward IPv4 link-local (`169.254.0.0/16`) traffic.
+Dropping it stops a sandbox guest reaching the host's
+cloud metadata service — and any other link-local endpoint. The
 destination drop goes in `raw PREROUTING` (host requests are `OUTPUT`, never
 `PREROUTING`, so the host keeps its own access); the source drop goes in
 `FORWARD`, **not** `PREROUTING`, so the host's own replies (e.g. an IMDS or
 `169.254.169.253` DNS response, delivered to `INPUT`) are left intact — a
-`raw PREROUTING -s` rule would drop them and break the host. IPv6 is not
-supported for sandbox guests, so it is disabled on guest interfaces and
-forwarded v6 is dropped:
+`raw PREROUTING -s` rule would drop them and break the host. 
+
+We also recommend disabling forwarding of IPv6 for VMs, unless you really know what you are doing.
 
 ```bash
 NIC=$(ip route show default | awk '{print $5}' | head -1)
--- a/README.md
+++ b/README.md
@@ -38,8 +38,9 @@ If you don't already have a Proxmox instance, see [CONTRIBUTING.md](CONTRIBUTING
 
 By default a sandbox VM can reach the Proxmox host's own services — the API
 (`pveproxy`, port 8006), SSH, etc. — via the SDN gateway, the `vmbr0` IP, or the
-host's external NIC. For cyber evals especially, you want those blocked so agents
-can't attack the Proxmox control plane.
+host's external NIC. Forwarded traffic can also reach cloud instance metadata
+services. For cyber evals especially, you want those blocked so agents can't
+attack the Proxmox or cloud control planes.
 
 This is **configured on the host at provisioning time, not by this library** — it
 needs the host's live routing table to know which interface external API/SSH
@@ -48,10 +49,20 @@ provisioning scripts in this repo (`scripts/virtualized_proxmox/build_proxmox_au
 and `scripts/ec2/userdata.sh`) set it up automatically, so hosts you create with
 them are isolated out of the box.
 
-If you provision Proxmox some other way, run these once **on the node**. They
-accept the management ports only on the default-route interface (where external
-callers arrive) and leave SDN DNS/DHCP open; sandbox VMs sit on other bridges and
-hit the default-deny policy:
+If you provision Proxmox some other way, configure equivalent persistent rules
+**on the node**. The Proxmox rules accept management ports only on the
+default-route interface (where external callers arrive) and leave SDN DNS/DHCP
+open. The remaining rules enforce RFC 3927 section 7, which Linux's forward path
+ignores: a router must not forward IPv4 link-local (`169.254.0.0/16`) traffic, but
+Linux forwards it anyway, so dropping it stops a sandbox guest reaching the host's
+cloud metadata service — and any other link-local endpoint — over its on-link route. The
+destination drop goes in `raw PREROUTING` (host requests are `OUTPUT`, never
+`PREROUTING`, so the host keeps its own access); the source drop goes in
+`FORWARD`, **not** `PREROUTING`, so the host's own replies (e.g. an IMDS or
+`169.254.169.253` DNS response, delivered to `INPUT`) are left intact — a
+`raw PREROUTING -s` rule would drop them and break the host. IPv6 is not
+supported for sandbox guests, so it is disabled on guest interfaces and
+forwarded v6 is dropped:
 
 ```bash
 NIC=$(ip route show default | awk '{print $5}' | head -1)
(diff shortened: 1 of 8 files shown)


src/proxmoxsandbox/scripts/ec2/launch.sh
```diff
@@ -66,9 +66,10 @@ RUN_ARGS=(
     --cpu-options "NestedVirtualization=enabled"
     --subnet-id "$SUBNET_ID"
     --security-group-ids "$SECURITY_GROUP_ID"
-    # Expose the instance's own tags (incl. Name) via IMDS so the OTel collector
-    # can label metrics with the Name (read at boot, no ec2:DescribeTags needed).
-    --metadata-options "InstanceMetadataTags=enabled"
+    # Require IMDSv2 and keep token responses on the host. A hop limit above 1
+    # allows nested sandbox VMs to retrieve host metadata credentials.
+    # InstanceMetadataTags lets the OTel collector read the instance Name.
+    --metadata-options "HttpTokens=required,HttpPutResponseHopLimit=1,HttpProtocolIpv6=disabled,InstanceMetadataTags=enabled"
```
bmillwood-aisi, line 72: my understanding is that `HttpPutResponseHopLimit=1` is the default, if that's right and we're nevertheless setting it here to prevent that from being overridden, I think it would be somewhat useful to say so in the comment

I guess the idea behind forcing IMDSv2 is because the hop limit is only effective there?

ideally also we'd say something about why we're disabling IPv6

src/proxmoxsandbox/scripts/ec2/README.md
```diff
@@ -87,7 +87,7 @@ aws ec2 run-instances --region "$REGION" \
     --cpu-options "NestedVirtualization=enabled" \
     --subnet-id "$SUBNET_ID" \
     --security-group-ids "$SECURITY_GROUP_ID" \
-    --metadata-options "InstanceMetadataTags=enabled"
+    --metadata-options "HttpTokens=required,HttpPutResponseHopLimit=1,HttpProtocolIpv6=disabled,InstanceMetadataTags=enabled"
```
bmillwood-aisi, line 90: would it make sense to reference `launch.sh` here instead of replicating it?

src/proxmoxsandbox/scripts/ec2/userdata.sh
```diff
@@ -363,6 +363,60 @@ RemainAfterExit=yes
...
+# Confine sandbox guests at the host's forwarding layer. Kept identical to the
+# on-first-boot heredoc in scripts/virtualized_proxmox/build_proxmox_auto.sh.
+cat > /usr/local/bin/inspect-proxmox-block-cloud-metadata.sh << 'BLOCK_METADATA'
+#!/bin/bash
+set -euo pipefail
+
+# Enforce RFC 3927: a router must not forward IPv4 link-local (169.254.0.0/16),
+# but Linux forwards it anyway, so a sandbox guest can otherwise reach the host's
+# metadata service over its on-link route. 169.254.169.254 covers
+# AWS/GCP/Azure/Oracle/DO; README notes Alibaba/Azure-WireServer outside the range.
```
bmillwood-aisi: I was briefly confused that this means the top-level README and not the one in this directory

src/proxmoxsandbox/scripts/ec2/userdata.sh
```diff
@@ -363,6 +363,60 @@ RemainAfterExit=yes
 WantedBy=multi-user.target
 FIXUP_NAT_UNIT
 
+# IPv6 is not supported for sandbox guests. SDN vnet bridges are created per
+# sample with generated names, so default.disable_ipv6 makes every interface
+# created after boot (i.e. the vnets) come up with no IPv6; the management NIC,
+# already up, keeps its own setting.
```
bmillwood-aisi, line 369: your testing section in your PR description doesn't mention anything about IPv6, have you confirmed this works?

src/proxmoxsandbox/scripts/ec2/userdata.sh
```diff
@@ -363,6 +363,60 @@ RemainAfterExit=yes
...
+cat > /etc/sysctl.d/99-inspect-proxmox-disable-ipv6.conf << 'SYSCTL_V6'
+net.ipv6.conf.default.disable_ipv6 = 1
+SYSCTL_V6
+
+# Confine sandbox guests at the host's forwarding layer. Kept identical to the
```
bmillwood-aisi, line 374: "kept identical" :(

I guess it would be a nuisance to factor this out so there was only one copy of it used in both places? if so, could we at least unit-test that they stay identical?

tests/proxmoxsandboxtest/test_host_isolation_e2e.py
```diff
@@ -76,6 +78,54 @@ async def test_sandbox_vm_cannot_reach_pveproxy_or_ssh() -> None:
...
         )
 
+        # A token PUT can time out solely because HttpPutResponseHopLimit=1,
+        # even when IMDS is reachable. A tokenless GET returns 401 when IMDSv2
+        # is required (or 200 when optional), so 000 specifically verifies that
+        # the host forwarding rule blocked the request.
+        metadata_get_res = await env.exec(
+            [
+                "curl",
+                "-sS",
```
bmillwood-aisi:
```suggestion
                "--silent",
                "--show-error",
```

tests/proxmoxsandboxtest/test_host_isolation_e2e.py
```diff
@@ -76,6 +78,54 @@ async def test_sandbox_vm_cannot_reach_pveproxy_or_ssh() -> None:
...
+        assert metadata_get_res.stdout.strip() == "000", (
+            "cloud instance metadata reachable from sandbox VM "
+            f"(curl returned http_code={metadata_get_res.stdout.strip()!r}). "
+            "Was the metadata forwarding block installed?"
+        )
+
+        metadata_token_res = await env.exec(
+            [
+                "curl",
+                "-sS",
```
bmillwood-aisi:
```suggestion
                "--silent",
                "--show-error",
```

tests/proxmoxsandboxtest/test_provisioning_security.py
```diff
@@ -0,0 +1,53 @@
...
+    REPO_ROOT
+    / "src"
+    / "proxmoxsandbox"
+    / "scripts"
+    / "virtualized_proxmox"
+    / "build_proxmox_auto.sh"
+)
+
+
+def test_ec2_launch_requires_imdsv2_with_one_hop() -> None:
```
bmillwood-aisi: this doesn't precisely describe what the test actually tests, I would say something like `contains_expected_imds_settings`

tests/proxmoxsandboxtest/test_provisioning_security.py
bmillwood-aisi, line 1: both the tests in this file are based on text in the provisioning scripts, not actual instance behaviour; if `test_host_isolation_e2e.py` were runnable in CI, I'd definitely advocate for getting rid of these completely, and though it isn't, I'm still a bit on the fence

tests/proxmoxsandboxtest/test_provisioning_security.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+@pytest.mark.parametrize(
+    "provisioner",
+    [
+        EC2_SCRIPTS / "userdata.sh",
+        VIRTUALIZED_SCRIPT,
+    ],
+)
+def test_provisioners_enforce_rfc3927_and_disable_ipv6(provisioner: Path) -> None:
```
bmillwood-aisi: again I'd say something more like `contain_expected_iptables_rules`

README.md
```diff
@@ -48,10 +49,20 @@ provisioning scripts in this repo (`scripts/virtualized_proxmox/build_proxmox_au
...
-If you provision Proxmox some other way, run these once **on the node**. They
-accept the management ports only on the default-route interface (where external
-callers arrive) and leave SDN DNS/DHCP open; sandbox VMs sit on other bridges and
-hit the default-deny policy:
+If you provision Proxmox some other way, configure equivalent persistent rules
+**on the node**. The Proxmox rules accept management ports only on the
+default-route interface (where external callers arrive) and leave SDN DNS/DHCP
+open. The remaining rules enforce RFC 3927 section 7, which Linux's forward path
+ignores: a router must not forward IPv4 link-local (`169.254.0.0/16`) traffic, but
+Linux forwards it anyway, so dropping it stops a sandbox guest reaching the host's
```
bmillwood-aisi: nit, but I think it's a little unfair to describe Linux as violating the spec here -- it's up to userspace to configure routing correctly I think?

README.md
```diff
@@ -62,8 +73,30 @@ pvesh create /nodes/$(hostname)/firewall/rules --type in --action ACCEPT --proto
 pvesh create /nodes/$(hostname)/firewall/rules --type in --action ACCEPT --proto udp --dport 67 --enable 1
 pvesh set /nodes/$(hostname)/firewall/options --enable 1
 pvesh set /cluster/firewall/options --enable 1
+iptables -w -t raw -C PREROUTING -d 169.254.0.0/16 -j DROP 2>/dev/null \
```
bmillwood-aisi, line 75: more duplication :(

bmillwood-aisi (commented):
I think none of my comments are blocking and this would be ok as-is, but take a look

src/proxmoxsandbox/scripts/ec2/userdata.sh
art-dsit: yes; I think it already didn't work before this PR, but just to be sure

src/proxmoxsandbox/scripts/ec2/userdata.sh
art-dsit: Unfortunately since other projects import this script directly there's no opportunity for a build step to synthesise a script from helpers.

This comment tends to be enough to make coding agents notice they need to edit the other one but I agree it's flaky.

If I wanted to fix this properly I would probably make a single python script which could be cleverer and figure out whether it's on ec2 vs nested metal and do the right thing.

tests/proxmoxsandboxtest/test_provisioning_security.py
art-dsit: ah yeah these are tests claude insists on adding, I have been trained to ignore them. I would much prefer proper e2e CI tests

src/proxmoxsandbox/scripts/ec2/README.md
art-dsit: It's a bit of a problem because this command is for launching after you baked an AMI (which I'm sure you realise); you don't really need these options in `launch.sh` at all (assuming you immediately bake the AMI rather than using the new template proxmox instance for an eval). So if anything this README is the primary source of truth

src/proxmoxsandbox/scripts/ec2/launch.sh
art-dsit: Yeah the readme says it's just a backstop. IMDSv1 is insecure and basically shouldn't be used

I've put a note in the top level README about IPv6

src/proxmoxsandbox/scripts/ec2/userdata.sh
art-dsit: I've removed it as it was weird slop

src/proxmoxsandbox/scripts/ec2/userdata.sh
bmillwood-aisi: > Unfortunately since other projects import this script directly there's no opportunity for a build step to synthesise a script from helpers.

I mean, we could build it and commit the result, right? Then other projects wouldn't be able to tell the difference, but we'd know what the source of truth was.

Happy not to push on this though.

art-dsit: Yeah we definitely could; the insane number of steps required to iterate on your proxmox eval would then increase by one, but as the author of the pipeline I'm not a good judge of whether that's likely to be a problem

bmillwood-aisi (approved):

Issue #81, opened 2026-06-15
exec/read_file output-size limit is enforced on encoded wire size, not decoded bytes


art-dsit:
SLOP ALERT

## Summary

`AsyncProxmoxAPI.read_file` enforces `max_size` against the **JSON-encoded wire bytes** of the `agent/file-read` response, not the decoded file/output size. With the default `decode=1`, non-ASCII content expands on the wire, so the *effective* limit shrinks with content and is **attacker-controllable** by code running in the sandbox. This is on the hot path — every `exec` reads `script.stdout`/`script.stderr` through it.

Deliberately split out from the UTF-8 decode fix (#80) so each lands cleanly.

## Why it happens

Proxmox returns file bytes with each raw byte mapped to a Latin-1 codepoint, re-encoded as UTF-8 JSON (see #80 / the `Wide character` note in `write_file`). Per output byte, the wire cost is:

| Source byte | On the wire | factor |
|---|---|---|
| printable ASCII (except `"` `\`) | itself | 1× |
| `"` `\`, `\n` `\r` `\t` `\b` `\f` | escaped | 2× |
| other control `0x00–0x1F` | `\u00XX` | **6×** |
| high byte `0x80–0xFF` | UTF-8 of U+0080–U+00FF | 2× |

So the advertised limit silently shrinks:

| output content | `MAX_EXEC_OUTPUT_SIZE` (10 MiB) actually trips at |
|---|---|
| ASCII text | ~10 MiB ✓ |
| high-byte binary (2×) | ~5 MiB |
| control-char-heavy / ANSI (up to 6×) | ~1.7 MiB |

Symptom: `OutputLimitExceededError` claiming "10 MiB exceeded" on output that was actually 1.7–5 MiB. Content-dependent, so it reads as flaky.

(Note: #78 found `agent/file-read` responses are gzipped, so the early `content-length > max_size` guard compares *compressed* size and rarely fires — the streaming byte-count path is the normal one. Either way the count is of encoded, not decoded, bytes.)

## Security flavor

In a cyber eval the sandbox runs untrusted code, which can deliberately emit control-char-heavy stdout to inflate the host-side buffer up to 6× and trip the limit early (or, if a `count` cap is added naively, blow client memory to 6× the intended bound). `decode=1` hands the in-sandbox payload a lever on the host's resource accounting.

## Proposed fix

[...]

PR #93, opened 2026-06-30
fix: large file/exec-output reads no longer crash with HTTP 597 Broken pipe

art-dsit:
## Problem

Reading a large or binary guest file (`read_file`) or command output (`exec`) fails with a non-standard **`HTTP 597 "Broken pipe"`**, retried 5× (deterministic, so every attempt fails), then propagates as an unhandled `HTTPStatusError` and crashes the caller.

## Root cause — pvedaemon self-aborts a large gzip'd response (isolated + instrumented on a live host)

Not body size per se, not the read, not `pveproxy` relaying — it's **response compression tripping a server-side timeout**:

| request (same 16 MiB random file) | result |
|---|---|
| `Accept-Encoding: gzip` (httpx default) | **597** |
| `Accept-Encoding: identity` | **200**, 34.5 MB body |
| 16 MiB *uniform* file, `gzip` | 200 (compresses to ~nothing) |

`pveproxy` forwards `Accept-Encoding: gzip` to `pvedaemon`, which builds the file-read response, then in `PVE::APIServer::AnyEvent::response()` **re-arms the client handle's 5 s inactivity timeout (`$self->{timeout}`) and only *then* runs `Compress::Zlib::memGzip` on the whole body**. For a large body `memGzip` blocks the event loop for >5 s, so when the loop resumes and `push_write`s the gzip'd body the timeout is **already overdue** and fires immediately.

Instrumented pvedaemon confirms it exactly:
```
RESP clen=22013635 gzip=1 → ABORT err=[Connection timed out] wbuf_remaining=19284451
```
The connection is torn down with 19.28 MB still unsent; only `22,013,635 − 19,284,451 = 2,729,184` bytes (one socket-buffer flush) reach the wire — a **truncated gzip stream** (I verified: `gunzip` yields "unexpected end of file"). `pvedaemon`'s `Content-Length` is *correct* (22 MB); the body is just cut short. Bumping that 5 s timeout to 60 s (nothing else changed) makes the same reads deliver the full body — intervention-level confirmation of the mechanism.

A lenient client (`curl`) takes the short body as `200` (partial, exit 18); **pveproxy's AnyEvent::HTTP sees body < Content-Length → 597**, which is what the sandbox client gets. The boundary is *sharp* (deterministic) because the thing being timed is CPU work (memGzip duration), not I/O — below ~5 s of memGzip the response delivers fully, above it truncates. `identity` skips memGzip, so it never blocks and delivers 34.5 MB fine.

This is an upstream Proxmox bug (arming a response timeout before a synchronous whole-body gzip). Proxmox documents only the 16 MiB *read* cap (→ truncation); this failure is undocumented.

## Fix

- **Send `Accept-Encoding: identity` on guest file reads** — opt out of pveproxy's response compression (the root cause). Even a 34.5 MB random body then transfers fine.
- **Move reads from `decode=1` to `decode=0`.** The default `decode=1` returns content as Latin-1-mangled UTF-8 (~2–6× on the wire, arriving corrupted and needing un-mangling). `decode=0` forwards base64 — **~4/3 the content size regardless of the bytes, and lossless.** Reads are bounded with `count`; oversized files/output surface as `OutputLimitExceededError` via Proxmox's `truncated` flag. PVE builds `decode=0` output by concatenating per-~1 MiB-chunk base64 (1 MiB isn't a multiple of 3, so segments carry their own `=` padding), so it's decoded segment-by-segment. `decode`/`count` are documented in the [PVE API viewer](https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/{node}/qemu/{vmid}/agent/file-read).

`decode=0` also **removes the Latin-1 mangling machinery** (`_recover_qga_bytes`, `_parse_truncated_file_read`, `_decode_exec_output`, the decode=1 streaming read) and the fragile truncation handling for cuts landing inside a multi-byte sequence — issues #77 / #80 are no longer reachable. Net **−137 lines** in `src/`.

## Tests (run against a real Proxmox instance)

- `test_file_read_597_is_response_compression` — characterises the upstream behaviour across `Accept-Encoding` headers: `identity` delivers a 14 MiB incompressible body (200); `gzip` and `gzip, deflate, br` return 596/597. Pins the root cause and guards the `identity` workaround.
- `test_exec_large_binary_output_surfaces_cleanly` — 30 MiB stdout → `OutputLimitExceededError`, not 597.
- `test_read_file_large_binary_surfaces_cleanly` — 14 MiB binary round-trips byte-exact (md5); 20 MiB → `OutputLimitExceededError`.
- Tightened `test_self_check` — the 50 MiB `test_read_and_write_large_file_binary` known-failure must now fail with `OutputLimitExceededError` specifically. It was previously excluded with an outcome-blind check (any non-pass accepted), which is why the suite never flagged the crash.
- `test_exec_10mb_limit` simplified — decode=0 reads whole bytes, so the even/odd byte-boundary parametrization is no longer needed.

Full `test_proxmox_sandbox_agent_commands.py` passes; `ruff` + `mypy` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fix: reading a large or binary guest file / command output no longer crashes with a non-standard HTTP 597 "Broken pipe". Root cause: `pveproxy` gzips the file-read response when the client accepts compression, and a large *incompressible* body (a real binary — compressible content shrinks and slips under the limit) dies mid-transfer with a 597 that escaped the retry layer. Guest file reads now send `Accept-Encoding: identity` to opt out of that compression. Reads also moved from the default `decode=1` (content as Latin-1-mangled UTF-8, ~2-6x on the wire, requiring un-mangling) to `decode=0` (base64, ~4/3 the content size and lossless) with a bounded `count`, so oversized files/output surface as `OutputLimitExceededError` and the Latin-1 mangling-recovery path is removed.
 - Fix: `exec()` no longer aborts the sample when a command kills its own command-runner wrapper process (e.g. `pkill -f`); it returns a failed `ExecResult` (`128+signal`, or `137` when the signal is unavailable) instead of raising a misleading `TimeoutError`
 
 ## 0.11.0 - 2026-06-01
--- a/tests/proxmoxsandboxtest/test_proxmox_sandbox_agent_commands.py
+++ b/tests/proxmoxsandboxtest/test_proxmox_sandbox_agent_commands.py
@@ -2,6 +2,7 @@ import hashlib
 from pathlib import Path
 from typing import List
 
+import httpx
 import pytest
 import tenacity
 from inspect_ai.util import OutputLimitExceededError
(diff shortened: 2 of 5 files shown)


art-dsit:
The alternative would be to patch proxmox (`.aisi2`) to bump the 5s time limit to something longer

dpolatajko-aisi (commented):
if i understand correctly, this PR incidentally also fixes #81 ? if so, it'd be nice to also include a unit test which covers that case (i.e. reading a file that would've expanded to break the limit before the change here)

otherwise this all looks good

art-dsit:
> if i understand correctly, this PR incidentally also fixes #81 ? if so, it'd be nice to also include a unit test which covers that case (i.e. reading a file that would've expanded to break the limit before the change here)
> 
> otherwise this all looks good

I believe the change to `large_file_result` in this PR does now cover the failure mode of #81 (was previously masked because the test didn't assert rigorously enough)

tests/proxmoxsandboxtest/test_proxmox_sandbox_agent_commands.py
```diff
@@ -38,49 +39,128 @@ async def test_exec_timeout_with_sigterm_handler_no_retryerror(
...
+    """read_file round-trips a multi-MiB binary and rejects oversized files.
+
+    14 MiB round-trips byte-exact (also exercises the decode=0 segment-decode);
+    20 MiB (over the 16 MiB cap) raises OutputLimitExceededError, not a 597.
+    """
+    if proxmox_sandbox_environment._is_windows():
+        pytest.skip("binary repro is Linux-only")
+    env = proxmox_sandbox_environment
+
+    # 14 MiB binary (multi-chunk, not 3-aligned) must round-trip byte-exact.
```
dpolatajko-aisi, line 94: >I believe the change to large_file_result in this PR does now cover the failure mode of https://github.com/UKGovernmentBEIS/inspect_proxmox_sandbox/issues/81

@art-dsit is this the round-trip that you are suggesting tests the bug in #81 ? i don't think the changes to `large_file_result` test that case, but I can see how this round-trip sort of does. it still seems a little conflated to me though - on main, this is going to hit the timeout because the random bytes take a long time to compress. but there's an edge case where you have a 14 MiB file of all the same (non-ascii) character, which compresses much faster, and the bug from #81 is isolated - rather than a 597 you get an erroneous `OutputLimitExceededError` because the wire size is >16MiB. i think it could be nice to isolate that test case, as documentation of the #81 bug that's been fixed

art-dsit: yeah sorry my comment was inaccurate (but I'll take refuge in the 'I believe' prefix)

Will add another test

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fix: reading a large or binary guest file / command output no longer crashes with HTTP 597 "Broken pipe".
```
dpolatajko-aisi, line 5: nit: possibly also worth mentioning the failure mode in #81 here

dpolatajko-aisi (approved):
nice !

PR #94, opened 2026-06-30
refactor: unify guest reads on decode=0, dropping the Latin-1 mangling path

art-dsit:
SLOP ALERT

**Stacked on #93** (base branch `fix/large-read-597-broken-pipe`) — review/merge that first; this diff is against it.

## What

Unify all guest reads — `read_file` and command stdout/stderr — onto a single agent file-read `decode=0` path, and delete the `decode=1` Latin-1 mangling that #93 worked around.

## Why

`decode=1` (the PVE API default the code inherited) reinterprets each file byte as a Latin-1 codepoint and serialises it as a JSON string. That:
- inflates the wire body — `0x80–0xFF` → 2 bytes each (~2× for binary), control bytes → `\uXXXX` 6 bytes each (up to 6×) — which is the root cause of the HTTP 597 in #93, and
- arrives mangled (`é` → `Ã©`), requiring `_recover_qga_bytes` to undo, plus fragile truncation handling for cuts landing mid-`\uXXXX` / mid-UTF-8-sequence (issues #77 / #80).

`decode=0` forwards the agent's base64 untouched: **~4/3 the content size regardless of the bytes, and lossless.** (The agent already base64-encodes internally, so `decode=1` was doing extra work to produce a *larger*, corrupted blob.) Params documented in the [PVE API viewer](https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/{node}/qemu/{vmid}/agent/file-read).

## Changes

- Removed: `async_proxmox.read_file` (decode=1 streaming), `_parse_truncated_file_read`, `_recover_qga_bytes`, `_decode_exec_output`, `agent_commands.read_file` / `read_file_or_blank`.
- Added: `agent_commands.read_file_capped_or_blank`.
- `read_file` and exec output reads both go through `read_file_capped` (decode=0); oversized reads surface as `OutputLimitExceededError` via the `truncated` flag.
- `test_exec_10mb_limit` drops its even/odd byte-boundary parametrization (decode=0 reads whole bytes, so a multi-byte sequence can't be split into a parse error — at most one trailing `U+FFFD`).

Net **−137 lines**. Removes the mojibake-recovery machinery, makes issues #77/#80 unreachable, and closes the pathological-content 597 edge that #93 left on the exec path (binary output is now bounded regardless of byte distribution).

## Tests (against a real Proxmox instance)

Full `test_proxmox_sandbox_agent_commands.py` passes — incl. `test_exec_10mb_limit` (emoji output truncates cleanly), the large-binary exec/`read_file` regressions, and `self_check`. `ruff` + `mypy` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fix: reading a guest file or command output larger than ~10 MiB no longer crashes with a non-standard HTTP 597 "Broken pipe". The agent file-read was issued without a `count` cap, so Proxmox tried to return the whole file in one base64-inflated JSON body that the `pveproxy`→`pvedaemon` hop could not deliver. `read_file()` now reads via the agent's `decode=0` mode (a compact, content-independent response body) so a full 16 MiB read stays deliverable and oversized files surface as `OutputLimitExceededError`; `exec()` output reads are capped at the output limit so they cannot overrun the proxy either.
 - Fix: `exec()` no longer aborts the sample when a command kills its own command-runner wrapper process (e.g. `pkill -f`); it returns a failed `ExecResult` (`128+signal`, or `137` when the signal is unavailable) instead of raising a misleading `TimeoutError`
 
 ## 0.11.0 - 2026-06-01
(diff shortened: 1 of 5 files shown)


art-dsit:
Folded into #93 as a single commit — not worth a separate PR. #93 now contains the full decode=0 change.

PR #95, opened 2026-07-03
test: centralize Proxmox pool/instance state reset in conftest

art-dsit:
## Problem

`QueueBasedProxmoxPool._instance_pools` and `InfraCommands._instances` are class-level, so they persist for the whole pytest session. `initialize()` is a no-op for a `pool_id` that already exists, so a `default` pool created by an e2e test (built from `PROXMOX_HOST` when `.env` is sourced) was silently reused by later mock tests that meant to build their own config from `PROXMOX_CONFIG_FILE`.

## Fix

A single autouse fixture in `conftest.py` that resets both registries before and after every test, and drop the now-redundant per-module `cleanup_infra_instances` fixtures and 28 scattered `clear_pools()` calls.

Every pool consumer is function-scoped and re-runs `task_init`, so resetting around each test is safe.


jasongwartz-aisi (approved):
nice improvement!

PR #96, opened 2026-07-04
GitHub CI (skips proxmox tests)

bmillwood-aisi:
Adds a workflow with a couple of jobs, which you can see below pass here.

Of course, all the best pytests are hard to run in this context. But we make do with the ones left over.

I'm trusting Claude not to over-mark tests as to be excluded here. I didn't check all of them would definitely fail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,26 @@
+on:
+  push:
+    branches:
+    - main
+  pull_request:
+
+jobs:
+  lint-and-typecheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v8
+
+      - name: Install dependencies
+        run: uv sync --locked
+
+      - name: Ruff lint
+        run: uv run ruff check .
+
+      - name: Ruff format
+        run: uv run ruff format --check .
+
+      - name: Type check
+        run: uv run mypy
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/checkout@v7
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v8
+        uses: astral-sh/setup-uv@v8.2.0
 
       - name: Install dependencies
         run: uv sync --locked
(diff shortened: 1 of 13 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,41 @@
...
+  push:
+    branches:
+    - main
+  pull_request:
+
+jobs:
+  lint-and-typecheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
```
art-dsit: can we please pin the version numbers with hashes

tests/proxmoxsandboxtest/test_built_in_vm.py
```diff
@@ -1,6 +1,10 @@
+import pytest
+
 from proxmoxsandbox._impl.built_in_vm import BuiltInVM
 from proxmoxsandbox._impl.qemu_commands import QemuCommands
 
+pytestmark = pytest.mark.proxmox
```
art-dsit: how about `req_proxmox` to match [req_k8s](https://github.com/UKGovernmentBEIS/inspect_k8s_sandbox/blob/main/test/k8s_sandbox/test_init_container.py#L15)

art-dsit (approved):
thanks!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,35 @@
...
+  push:
+    branches:
+    - main
+  pull_request:
+
+jobs:
+  lint-and-typecheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
```
bmillwood-aisi, line 11: nb that this comment is not forced to be in sync with the hash in any way :(

uv.lock
bmillwood-aisi, line 1: I don't remember intentionally making this change but I think it does need to be here (pyproject.toml doesn't set this, so `uv sync --locked` would object that the lockfile is out of sync; the alternative would be adding it to the pyproject file)

bmillwood-aisi:
@art-dsit I applied your suggestions but then did a couple of other things as well (e.g. actionlint), so I decided to re-request review rather than just merge. feel free to merge if you're happy

.github/workflows/ci.yml
art-dsit: yup but hopefully only AI agents will be updating this, and they'll remember to do the comment

PR #97, opened 2026-07-06
misc refactoring

bmillwood-aisi:
Three commits that I think helped me understand what was going on while I was making another change. They're probably easiest to review commit by commit.

I might have chosen bad names by virtue of not knowing the codebase that well. Feel free to suggest tweaks.

# Testing

I didn't test these changes in isolation, but I tested #98 which is stacked on top of this.

### DIFF
--- a/src/proxmoxsandbox/_impl/qemu_commands.py
+++ b/src/proxmoxsandbox/_impl/qemu_commands.py
@@ -8,7 +8,6 @@ from typing import Collection, Dict, List, Set
 import httpx
 import tenacity
 from inspect_ai.util import trace_action
-from pydantic.networks import HttpUrl
 
 from proxmoxsandbox._impl.async_proxmox import (
     AsyncProxmoxAPI,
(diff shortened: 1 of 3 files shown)


src/proxmoxsandbox/_impl/qemu_commands.py
```diff
@@ -230,6 +228,9 @@ async def create_and_start_vm(
                 "os_type is only supported for OVA or existing_vm_template_tag"
             )
```
bmillwood-aisi, line 229: I don't fully understand the significance of these conditions, but I wonder if:

1. they should be asserted in the `vm_config` validation? (is it that these configurations are only not supported *here*, or are they nonsensical in general?)
2. we should allow e.g. `vm_config.disk_controller` to be specified in the non-`ova` case, but in that case it validates against the disk controller rather than creating it (this means that you can easily switch out the `vm_source_config` and know that your `disk_controller` setting is not going to embed any surprises)

src/proxmoxsandbox/_impl/qemu_commands.py
```diff
@@ -230,6 +228,9 @@ async def create_and_start_vm(
...
 
+        vm_id_to_clone: int
+        preserve_tags: bool
```
bmillwood-aisi, line 232: these aren't strictly necessary and are more aids to the reader than anything else

src/proxmoxsandbox/_impl/qemu_commands.py
```diff
@@ -230,175 +255,121 @@ async def create_and_start_vm(
...
-                    "couldn't find template VM for "
-                    + f"{vm_config.vm_source_config.built_in}"
-                )
-
-            new_vm_id = await self.clone_vm_and_start(
-                vm_config, vm_id_to_clone, sdn_vnet_aliases, False
-            )
+            preserve_tags = False
         elif vm_config.vm_source_config.ova is not None:
-            if isinstance(vm_config.vm_source_config.ova, HttpUrl):
```
bmillwood-aisi, line 246: read this part with whitespace ignored (it's in the cog button next to Submit review)

src/proxmoxsandbox/_impl/qemu_commands.py
art-dsit: I think these conditions are only here because of incomplete implementation - would be best to finish the implementation and remove them; deeper in the code it would make sense to validate against incorrect config such as a built-in debian vm with os_type = windows

src/proxmoxsandbox/_impl/qemu_commands.py
```diff
@@ -230,175 +255,121 @@ async def create_and_start_vm(
...
+
+        new_vm_id = await self.clone_vm_and_start(
+            vm_config=vm_config,
+            vm_id_to_clone=vm_id_to_clone,
+            sdn_vnet_aliases=sdn_vnet_aliases,
+            preserve_tags=preserve_tags,
+        )
+
         if new_vm_id is None:
             raise ValueError("No VM ID?")
```
art-dsit, line 374: I believe this guard can also go, although it is python so maybe keep it if we don't trust the type hints

src/proxmoxsandbox/_impl/qemu_commands.py
art-dsit: the exact `if` statement format used to make more sense when there were other, since dropped, types of vm_config

art-dsit (approved):
LGTM, thanks it's much clearer now. Will merge if the test suite passes

Issue #69, opened 2026-05-05
OVA template cache key omits VmConfig settings (disk_controller / cpu / os_type) — silently stale templates after sandbox.py edits


sinman-aisi:
## Summary

When a VM is imported from an OVA, `proxmoxsandbox` creates a Proxmox-native VM template tagged by the OVA filename + size (`ova-<filename>_ova-<bytes>`). Subsequent runs reuse that template via `find_existing_template_with_tag`. The tag is keyed on **OVA identity only** — not on the `VmConfig` knobs that affect the imported template's controllers (`disk_controller`, `cpu`, `os_type`, possibly others). So if you change one of those settings in `sandbox.py` after a template has already been created, the change is silently ignored on every later run.

## How it bit us

I have a 5-VM Proxmox sandbox; pfSense is one of the targets and runs FreeBSD 12.2 inside the OVA. Under nested virt with the default `disk_controller="scsi"` (Proxmox attaches as `scsi0` on a `virtio-scsi-single` controller), pfSense's kernel doesn't have a `vtscsi` driver and boot-loops endlessly on:

```
Trying to mount root from ufs:/dev/ufsid/675ad6c6fba4e884 [rw,noatime]...
Root mount waiting for: CAM usbus0
Root mount waiting for: CAM usbus0
...
```

The fix in `sandbox.py` is to set `disk_controller="ide"` on pfSense's `VmConfig`. Easy. The footgun is that **if you add this fix to a `sandbox.py` whose pfSense template was previously imported with the default `scsi`, the new setting has no effect** — `find_existing_template_with_tag` finds the old template and clones from it. The cloned VM has `scsi0` and pfSense boot-loops as before. Symptom: `mean=0.0` across all samples with deterministic "No route to host" / target-VM-unreachable errors that look like networking bugs but are actually the target VM never finishing boot.

The mitigation we used was to delete the cached templates per-instance and let the next run re-import. But that's only obvious *after* you've root-caused the wrong thing several times.

## Repro

```python
# First run — pfSense VmConfig with no disk_controller; default "scsi" applied.
VmConfig(
    name="pfSense",
    vm_source_config=VmSourceConfig(ova=Path("vm-200.ova")),
    nics=(...),
    is_sandbox=False,
)
# Run, OVA imports as scsi0, template cached.

# Edit sandbox.py to add disk_controller="ide" (because BSD-on-nested-virt needs IDE).
VmConfig(
    name="pfSense",
[...]

PR #99, opened 2026-07-11
Verify disk_controller against clone source for non-OVA VMs

bmillwood-aisi:
The motivation here is that I want to write my `disk_controller` into my `VmConfig` and not have to comment it out when I flip between `existing_vm_template_tag` and `ova`. But at the same time, `existing_vm_template_tag` can't *change* the `disk_controller` of the template, and I don't want a `VmConfig` that says `disk_controller` that doesn't produce a VM with that controller.

The solution is to validate the `disk_controller`, so that I can look at a config and know that because it works it's correct, and swapping in an equivalent `ova` won't burn me.

I don't know if this is worth the complexity? I think conceptually it makes sense but I don't *need* it.

I've run the `req_proxmox` tests manually and they seem fine.

### DIFF
--- a/src/proxmoxsandbox/_impl/qemu_commands.py
+++ b/src/proxmoxsandbox/_impl/qemu_commands.py
@@ -2,7 +2,7 @@ import abc
 import re
 import tarfile
 from logging import getLogger
-from typing import Collection, Dict, List, Set
+from typing import Collection, Dict, List, Literal, Set
 
 import httpx
 import tenacity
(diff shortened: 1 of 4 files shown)


art-dsit:
Happy to merge this but we really ought to also attend to #69

art-dsit (approved):

bmillwood-aisi:
> Happy to merge this but we really ought to also attend to #69

thanks for the flag -- isn't this PR doing the #3 thing there? granted that #2 is still worth doing even so

PR #100, opened 2026-07-14
Log acquired pool instance at INFO for host attribution (captured in the .eval)

art-dsit:
With an instance pool, the `sandbox.config.host/port/node` recorded in each sample's eval log are whatever the env vars (or the eval's explicit config) said at construction — never the instance the sample acquired. `ProxmoxSandboxEnvironmentConfig` is frozen and defaults those fields from process env vars, so with a pool they can't attribute a sample to a server. An eval log showed every sample attributed to the same host while QGA retry warnings proved they were running on different servers.

`sample_init` now logs the acquired instance at INFO, immediately after acquisition (before anything can fail):

```
Acquired instance <id> from pool '<pool>': host=<host> port=<port> node=<node>
```

## Making INFO actually land in the `.eval`

Inspect's transcript captures `info`+ — but only for loggers it configures. `init_logger` sets the **root** logger to the `warning` default and lowers only `inspect_ai`'s own package loggers; a registered extension like this one inherits root=`warning`, so `logger.info(...)` is filtered at the source and never reaches the transcript. (Confirmed empirically: a default `inspect eval` captured zero logger events; the line only appeared when `--log-level info` was passed.)

So the provider opts its own logger in at import: `getLogger("proxmoxsandbox").setLevel(INFO)`. This is downstream-overridable — `--log-level-transcript warning` (or `INSPECT_LOG_LEVEL_TRANSCRIPT=warning`) still drops it, and the console stays quiet by default (the display gate is `--log-level`, still `warning`).

Do not use the serialised `sandbox.config` host/port/node for attribution in pool mode — the README says so.

## Verification

Real `inspect eval` against a Proxmox at `172.31.106.106`, three log-level configs:

| config | in `.eval` transcript | console |
| --- | --- | --- |
| default | present | quiet |
| `--log-level-transcript warning` | absent | quiet |
| `--log-level info` | present | present |

A full `ctf4` run (`message_limit=20`, 42 model turns / 42 tool calls, 248 transcript events) produced 15 INFO lines total — all one-time VM lifecycle (acquire / create / ready / QGA ping / cleanup / release); the count does not scale with agent activity. The sentinel and concurrent-distinct-instances cases are covered by unit tests in `test_host_attribution.py`, which fail without the fix.

Follow-ups: consider excluding the credential fields from the serialised config entirely, since in pool mode they are pure misdirection; cyber-register's `_load_proxmox_creds` (which stamps `instances[0]` creds into the config, producing the real-looking wrong host in the incident log) can likewise stop setting them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,8 @@ export PROXMOX_CONFIG_FILE=/path/to/instances.json
 
 Instances with the same `pool_id` form a pool. Each eval sample acquires one instance from its pool, uses it exclusively, and releases it back when done. Concurrency is automatically limited to the total number of instances.
 
+Each sample records the instance it acquired in the Inspect sample store, under the keys `proxmox:instance_id`, `proxmox:host`, `proxmox:port`, `proxmox:node` and `proxmox:pool_id` — check `sample.store` in the `.eval` log to attribute a sample to a Proxmox server. Do not use the `host`/`port`/`node` fields of the sample's serialised sandbox config for this: with a pool those fields are not connected to the instance the sample ran on.
+
 ## Configuring
 
 Here is a full example sandbox configuration. 
--- a/README.md
+++ b/README.md
@@ -151,8 +151,6 @@ export PROXMOX_CONFIG_FILE=/path/to/instances.json
 
 Instances with the same `pool_id` form a pool. Each eval sample acquires one instance from its pool, uses it exclusively, and releases it back when done. Concurrency is automatically limited to the total number of instances.
 
-Each sample logs the instance it acquired at `INFO` level (`Acquired instance <id> from pool '<pool>': host=<host> port=<port> node=<node>`), which lands in the sample transcript of the `.eval` log — check it to attribute a sample to a Proxmox server. Do not use the `host`/`port`/`node` fields of the sample's serialised sandbox config for this: with a pool those fields are not connected to the instance the sample ran on. To suppress these lines, run with `--log-level-transcript warning`.
-
 ## Configuring
 
 Here is a full example sandbox configuration. 
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ export PROXMOX_CONFIG_FILE=/path/to/instances.json
 
 Instances with the same `pool_id` form a pool. Each eval sample acquires one instance from its pool, uses it exclusively, and releases it back when done. Concurrency is automatically limited to the total number of instances.
 
-Each sample records the instance it acquired in the Inspect sample store, under the keys `proxmox:instance_id`, `proxmox:host`, `proxmox:port`, `proxmox:node` and `proxmox:pool_id` — check `sample.store` in the `.eval` log to attribute a sample to a Proxmox server. Do not use the `host`/`port`/`node` fields of the sample's serialised sandbox config for this: with a pool those fields are not connected to the instance the sample ran on.
+Each sample logs the instance it acquired at `INFO` level (`Acquired instance <id> from pool '<pool>': host=<host> port=<port> node=<node>`) — check the logs to attribute a sample to a Proxmox server. Do not use the `host`/`port`/`node` fields of the sample's serialised sandbox config for this: with a pool those fields are not connected to the instance the sample ran on.
 
 ## Configuring
 
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ export PROXMOX_CONFIG_FILE=/path/to/instances.json
 
 Instances with the same `pool_id` form a pool. Each eval sample acquires one instance from its pool, uses it exclusively, and releases it back when done. Concurrency is automatically limited to the total number of instances.
 
-Each sample logs the instance it acquired at `INFO` level (`Acquired instance <id> from pool '<pool>': host=<host> port=<port> node=<node>`) — check the logs to attribute a sample to a Proxmox server. Do not use the `host`/`port`/`node` fields of the sample's serialised sandbox config for this: with a pool those fields are not connected to the instance the sample ran on.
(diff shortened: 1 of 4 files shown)


art-dsit:
_Comment by Claude, recording design discussion so it isn't lost._

Art's review question: why write to the sample store rather than recording the sandbox config the way other providers do, so it shows up under Sample Info in inspect view?

Findings from reading inspect_ai (0.3.239 and current main):

- `EvalSample.sandbox` is copied verbatim from the *requested* spec (`_eval/task/run.py`, `sandbox=sample.sandbox`) before the sample runs, and is only populated for sample-level specs (task-level sandboxes leave it None; viewers fall back to the eval header). No provider output ever reaches it: `sample_init` returns only the live environments dict, and neither installed nor upstream inspect_ai has a provider→`sample.sandbox` channel.
- Other providers don't record runtime identity either (a k8s sample doesn't log which node/namespace it landed on) — their configs merely happen to be truthful because request fully determines placement. Pooled Proxmox binds sample→server at runtime, so the same mechanism becomes misattribution instead of absence.
- The store is the one durable per-sample surface a provider can write today (`init_subtask_store` runs before `sandboxenv_context`, so `store()` inside `sample_init` resolves to the sample's store, serialised into `EvalSample.store`).

The idiomatic fix would be an upstream inspect_ai change: an optional `SandboxEnvironment` classmethod (e.g. `sample_resolved_config(environments)`) called after `init_sandbox_environments_sample`, whose non-None result replaces `sample.sandbox` with a resolved `SandboxEnvironmentSpec` — the log and inspect view's Sample Info then pick it up with no further changes. This provider would return `config.model_copy(update={"host": instance.host, "port": ..., "node": ...})`. Wrinkle needing maintainer sign-off: `EvalSample.sandbox` changes meaning from "requested" to "realized" whenever a provider reports.

Decision for now: keep the store keys (works with today's inspect_ai); revisit the upstream hook separately.

README.md
```diff
@@ -151,6 +151,8 @@ export PROXMOX_CONFIG_FILE=/path/to/instances.json
 
 Instances with the same `pool_id` form a pool. Each eval sample acquires one instance from its pool, uses it exclusively, and releases it back when done. Concurrency is automatically limited to the total number of instances.
 
+Each sample records the instance it acquired in the Inspect sample store, under the keys `proxmox:instance_id`, `proxmox:host`, `proxmox:port`, `proxmox:node` and `proxmox:pool_id` — check `sample.store` in the `.eval` log to attribute a sample to a Proxmox server. Do not use the `host`/`port`/`node` fields of the sample's serialised sandbox config for this: with a pool those fields are not connected to the instance the sample ran on.
```
bmillwood-aisi:
```suggestion
Each sample records the instance it acquired in the Inspect sample store, under the keys `proxmox:instance_id`, `proxmox:host`, `proxmox:port`, `proxmox:node` and `proxmox:pool_id` — check `sample.store` in the `.eval` log to attribute a sample to a Proxmox server.
```

it's not clear to me whether those fields are naturally there or whether that's just an artifact of the way we use the library, and either way we're going to fix it so I wouldn't bother documenting it here

bmillwood-aisi (commented):
I'm not sure this is "right" but not having this information is so annoying that I'm happy to go with this until we explore something else

though as mentioned in Slack maybe we should just go with a log message?

art-dsit:
As discussed we'll just go with a log message (I suppose at `info`) instead

src/proxmoxsandbox/_proxmox_sandbox_environment.py
```diff
@@ -37,6 +37,12 @@
     ProxmoxSandboxEnvironmentConfig,
 )
 
+# Inspect only lowers its own package loggers below the root WARNING default, so a
+# third-party provider's INFO never reaches the .eval transcript. As an Inspect
+# extension we opt our own logger in; a user can still silence it with
+# --log-level-transcript warning (that gate is downstream of this level).
+getLogger("proxmoxsandbox").setLevel(INFO)
+
```
bmillwood-aisi, line 48: I asked Fable about this. It defended the use of top-level side effects for this case, but added these points:

```
1. Scope is wider than the PR's stated purpose. getLogger("proxmoxsandbox") is the parent of every 
getLogger(__name__) in the package, so this opts ~21 existing INFO sites into transcripts (VM creation 
progress, template lookup, SDN setup…), not just the host-attribution line. That's probably fine — they're 
low-volume per-sample setup logs and arguably useful transcript context — but it's a real behavior 
change and the CHANGELOG entry ("Opt logger into INFO level") only gestures at it. For comparison, 
inspect_k8s_sandbox doesn't do this, so its INFO is lost at defaults; you're breaking new ground rather 
than following ecosystem convention.
2. A politeness guard is available if you want it: if getLogger("proxmoxsandbox").level == NOTSET: 
...setLevel(INFO). That preserves the fix at defaults while never overriding a host application that 
configured the level before the entry point gets imported. Minor, but it makes the top-level effect strictly 
non-destructive.
```

Seems fine to me, but highlighting in case you want to check the effect on a run or make some change.

bmillwood-aisi (approved):

tests/proxmoxsandboxtest/test_host_attribution.py
bmillwood-aisi, line 1: I did not read this closely. It seems probably fine.

src/proxmoxsandbox/_proxmox_sandbox_environment.py
art-dsit: Yeah I'm not going to add every single thing to the changelog (these AI agents don't have a good conception of humans' TLDR propensity), but I like 2, in fact if we want the user to able to set our .level to DEBUG it's the only way to avoid overwriting that I think?

tests/proxmoxsandboxtest/test_host_attribution.py
art-dsit: ditto

PR #103, opened 2026-07-20
Drop misleading single-instance fields from ProxmoxSandboxEnvironmentConfig

bmillwood-aisi:
As you'll see in the changes, these aren't *entirely* unused, but the remaining uses were all misleading and/or easily replaced.

The config compatibility story is surprisingly good. We stop serialising the dropped fields. Because they all had defaults, old code keeps being able to read new serialisations. Because pydantic ignores unknown fields by default, new code continues to be able to read old serialisations (but ignores the values in the extra fields, so they won't roundtrip).

# Testing

Unit tests pass. I haven't yet had time to run either the `req_proxmox` tests nor run a real range against this and confirm that doesn't break either.

### DIFF
--- a/tests/proxmoxsandboxtest/test_credential_redaction.py
+++ b/tests/proxmoxsandboxtest/test_credential_redaction.py
@@ -7,13 +7,11 @@ import pytest
 from pydantic import SecretStr, ValidationError
 
 from proxmoxsandbox._impl.async_proxmox import AsyncProxmoxAPI
-from proxmoxsandbox._impl.infra_commands import InfraCommands
 from proxmoxsandbox._proxmox_sandbox_environment import (
     ProxmoxSandboxEnvironment,
 )
 from proxmoxsandbox.schema import (
     ProxmoxInstanceConfig,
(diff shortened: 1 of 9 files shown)


tests/proxmoxsandboxtest/test_credential_redaction.py
```diff
@@ -107,37 +102,3 @@ async def test_cleanup_failure_log_excludes_instance_password(caplog):
     assert "host=127.0.0.1" in messages
     assert "port=8006" in messages
     assert "node=proxmox" in messages
-
-
-@pytest.mark.asyncio
-async def test_task_cleanup_debug_log_excludes_config_password(caplog):
```
bmillwood-aisi, line 113: maybe removing this test entirely was overly aggressive, but it doesn't make as much sense now that `ProxmoxSandboxEnvironmentConfig` never includes a password

art-dsit:
Tests pass (except test_sandbox_vm_cannot_reach_host_or_cloud_metadata but that's an artifact of our nitrovirt ami needing to be rebuilt I think)

art-dsit (approved):

PR #106, opened 2026-07-27
bump _QGA_MAX_RETRIES

bmillwood-aisi:
We've seen quite a bit of "500 QEMU guest agent is not running" failures. Crucially, sometimes these recover by themselves, sometimes they exhaust their retries and die. In the modern era of the typical sample taking hours to days, retrying for less than 60 seconds before giving up on the sample seems not tenacious enough to me.

Since each attempt waits for 3 seconds (as far as I know we don't control this) and the retry waits double from 2 up to a max of 20, 5 attempts is approximately 3 + 2 + 3 + 4 + 3 + 8 + 3 + 16 + 3 = 45 seconds, and 25 attempts is that plus 20 more attempts * (20 + 3) per attempt = 505 seconds, or 8m25s.

You can imagine doing more sophisticated things here as well, but let's get this out the door and see how it is in practice.

I have not tested this. It's very simple and testing it "properly" would mean deliberately inducing a failure, which might be annoying to do.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Fix: reading a large or binary guest file / command output no longer crashes with HTTP 597 "Broken pipe" (on Proxmox >= 9.2).
 - Security: redact Proxmox passwords in configuration representations and validation error messages, and omit credentials from cleanup logs
 - Fix: `exec()` no longer aborts the sample when a command kills its own command-runner wrapper process
+- Fix: "500 QEMU guest agent is not running" is retried for much longer (~45s -> 8m25s)
 
 ## 0.11.0 - 2026-06-01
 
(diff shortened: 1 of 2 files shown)


dpolatajko-aisi (approved):

ekin-aisi:
Excited for this to go in, hitting the same issue. I think the longer term fix is to bump CPU of the attacker kali VM?

bmillwood-aisi:
> Excited for this to go in, hitting the same issue. I think the longer term fix is to bump CPU of the attacker kali VM?

I think there's no amount of CPU we can give the kali VM that the agent can't exhaust if it wants to. I would guess if there's a longer-term fix it might be something like, run the QGA at a higher priority, or limit the amount of resources the agent can consume on the rest of the box so there's always room left for the QGA

PR #107, opened 2026-07-28
feat: add a network egress lockdown bootstrap script

dpolatajko-aisi:
## Description

Adds a new bootstrap script for preventing guest VM network access via requests forward by the host. This bootstrap script does not run by default, but can be configured to do so.

## Limitations

This approach does not prevent the host from making outbound network calls. Since the host runs a `dnsmasq` proxy, outbound traffic from the agent is technically still possible, via DNS resolution.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,6 +77,21 @@ export PROXMOX_WINDOWS_TEMPLATE_TAG=<your-tag>
 
 With this set, tests in `test_proxmox_sandbox_agent_commands.py` will run for both Linux and Windows.
 
+### Egress lockdown test
+
+`test_egress_lockdown_e2e.py` is skipped unless `PROXMOX_EGRESS_LOCKDOWN_ENABLED`
+is set, because it needs the host's egress lockdown active — and a locked-down
+host fails other integration tests that expect guests to have egress. To run it:
+
+```bash
+ssh root@<proxmox-host> 'touch /etc/inspect-proxmox-egress-lockdown && systemctl start inspect-proxmox-egress-lockdown.service'
+PROXMOX_EGRESS_LOCKDOWN_ENABLED=1 uv run pytest tests/proxmoxsandboxtest/test_egress_lockdown_e2e.py
+ssh root@<proxmox-host> 'rm /etc/inspect-proxmox-egress-lockdown && systemctl start inspect-proxmox-egress-lockdown.service'
+```
+
+Remember the last step: leaving the marker in place breaks the rest of the
+integration suite.
+
 ### Debug logging
 
 To see debug-level log output while running tests:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,7 +81,13 @@ With this set, tests in `test_proxmox_sandbox_agent_commands.py` will run for bo
 
 `test_egress_lockdown_e2e.py` is skipped unless `PROXMOX_EGRESS_LOCKDOWN_ENABLED`
 is set, because it needs the host's egress lockdown active — and a locked-down
-host fails other integration tests that expect guests to have egress. To run it:
+host fails other integration tests that expect guests to have egress.
+
+The built-in VM template must already exist on the host before locking down:
+baking it boots the source image and installs packages (including
+`qemu-guest-agent`) from inside the guest, which needs guest egress. Any test
+that creates a default sandbox (e.g. `test_host_isolation_e2e.py`) bakes it on
+first use. Then:
 
 ```bash
 ssh root@<proxmox-host> 'touch /etc/inspect-proxmox-egress-lockdown && systemctl start inspect-proxmox-egress-lockdown.service'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - Log the acquired pool instance (`host`/`port`/`node`) at `INFO`
 - Opt logger into `INFO` level for consistency with Inspect core
 - Prevent VMs from accessing cloud instance metadata credentials, disable IPv6 for sandbox guests (when using the bundled provisioning scripts)
-- Optional marker-file-gated egress lockdown for sandbox guests, installed inert by the bundled provisioning scripts (no behaviour change unless `/etc/inspect-proxmox-egress-lockdown` is created); when active it drops guest traffic forwarded via all default-route interfaces, reasserts the rules every minute via a systemd timer, fails closed (blanket forward drop + Proxmox API halt on failure), and stops the SDN `dnsmasq` instances from recursing upstream — backstopped by a firewall rule dropping upstream `dnsmasq` traffic — closing the DNS-tunnel channel
+- Optional egress lockdown for sandbox guests (when using the bundled provisioning scripts): drops forwarded guest traffic and stops the SDN `dnsmasq` instances recursing upstream. Installed inert, so nothing changes until `/etc/inspect-proxmox-egress-lockdown` is created; see the README
 - Fix: guest file / command-output reads work again on Proxmox < 9.2.
 - Fix: reading a large or binary guest file / command output no longer crashes with HTTP 597 "Broken pipe" (on Proxmox >= 9.2).
 - Security: redact Proxmox passwords in configuration representations and validation error messages, and omit credentials from cleanup logs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - Log the acquired pool instance (`host`/`port`/`node`) at `INFO`
 - Opt logger into `INFO` level for consistency with Inspect core
 - Prevent VMs from accessing cloud instance metadata credentials, disable IPv6 for sandbox guests (when using the bundled provisioning scripts)
-- Optional marker-file-gated egress lockdown for sandbox guests, installed inert by the bundled provisioning scripts (no behaviour change unless `/etc/inspect-proxmox-egress-lockdown` is created)
+- Optional marker-file-gated egress lockdown for sandbox guests, installed inert by the bundled provisioning scripts (no behaviour change unless `/etc/inspect-proxmox-egress-lockdown` is created); when active it drops forwarded guest traffic and stops the SDN `dnsmasq` instances from recursing upstream, closing the DNS-tunnel channel
 - Fix: guest file / command-output reads work again on Proxmox < 9.2.
 - Fix: reading a large or binary guest file / command output no longer crashes with HTTP 597 "Broken pipe" (on Proxmox >= 9.2).
 - Security: redact Proxmox passwords in configuration representations and validation error messages, and omit credentials from cleanup logs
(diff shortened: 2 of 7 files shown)


jon-aisi (commented):
cool, looks good - would like to make this somewhat more robust but what's there seems good

thoughts after chatting with the bots:
1. no verification that these rules have been applied: can we fail noisily and kill the host / proxmox if these restrictions fail to apply?
2. rules can be reverted by subsequent processes: would be great to have them reasserted might be very hard given (iiuc) proxmox runs as root
3. no tests asserting this script has the desired behaviour: it's complex enough I don't trust just reviewing code to catch everything
4. only restricts default host nic (a second one would be skipped)
    - and more strongly it seems like this design assumes default SDN topology and variants could silently expose bad behaviour we don't want. can we assert the expected topology if this restriction is enabled?


GPT:

>  1. Keep the firewall rule at the very front
> 
>   Firewall rules are checked in order. If a later service adds an “allow traffic” rule above this lockdown rule, traffic could pass
>   before it reaches the “drop” rule.
> 
>   Change the script so every run does:
> 
>   1. delete its old lockdown rules
>   2. insert fresh lockdown rules at rule position 1
> 
>   That makes restarting the service repair rule order.
> 
>   2. Block DNS with firewall too, not only by editing a file
> 
>   Right now DNS egress is stopped by blanking /run/dnsmasq/resolv.conf. That is useful, but it depends on dnsmasq continuing to use that
>   file exactly as expected.
> 
>   I’d add a second protection: firewall-block outbound DNS from the Proxmox-hosted dnsmasq processes to anything outside the sandbox
>   networks. In plain terms:
> 
>   - guests can still ask Proxmox DNS for internal VM names
>   - Proxmox DNS is not allowed to ask the internet/VPC resolver for outside names
>   - if the resolver file gets recreated, DNS egress is still blocked
> 
>   3. Say or enforce which network layout this supports
> 
>   The current rule blocks traffic crossing the host’s default internet-facing NIC. That is right for the bundled EC2/simple-SDN setup.
> 
>   But it does not automatically cover every possible Proxmox network design, for example:
> 
>   - a VM attached to an existing bridge someone made manually
>   - a second host NIC
>   - a VPN/tunnel interface
>   - a template VM whose existing NICs are left untouched
> 
>   So I’d either document clearly:
> 
>   > This lockdown is only a strong control for the bundled generated SDN topology.
> 
>   Or better, enforce it:
> 
>   > When lockdown is enabled, reject or avoid VM configs using existing VNETs/template NICs unless explicitly allowed.
> 
>   4. Add an end-to-end test
> 
>   Add a test that turns the marker on, starts a sandbox VM, then checks:
> 
>   - curl https://example.com fails
>   - curl http://1.1.1.1 fails
>   - DNS lookup for an outside name fails
>   - internal VM-to-VM traffic still works if expected
>   - host package/cloud-agent access still works, or at least is not affected by the guest rules

Claude:

> Here's the honest gap list, grouped by why each one exists. Keeping to the host/Proxmox layer as asked.
> 
>   By design — out of scope for this mechanism
> 
>   - The host's own egress. Anything running as the host (root shell, Proxmox services, the CloudWatch agent) has full internet. The PR
>   only constrains the FORWARD path. This is the stated limitation and it's the right call for this layer — but it means every gap below
>   that ends in "attacker gets host code exec" ends in "full egress."
>   - Guest↔guest traffic across vnets. Forwarding between two vnet bridges never touches the mgmt NIC, so the rules don't see it. Fine
>   today because it's one sample per instance, so all guests on a host belong to the same eval — but it's an assumption worth keeping in
>   mind if that invariant ever changes.
>   - Inbound from the VPC to the host (API, SSH). pve-firewall limits it to 8006/22 on the mgmt NIC, but who can reach those ports is
>   inherently the SG's job, not this PR's.
> 
>   Reachable attack surface that can defeat it
> 
>   - dnsmasq is still guest-reachable on 53 and 67. The lockdown removes its upstream, but guests can still throw arbitrary DNS and DHCP
>   packets at its parsers. dnsmasq runs as root's child on the host (dropping to the dnsmasq user, but still on the host), and it has a
>   history of memory-corruption CVEs in exactly these code paths (DNSpooq, 2021). An exploit here lands inside the boundary.
>   - QEMU/virtio and the guest agent channel. The VM escape surface — emulated devices, virtio queues, the QGA virtio-serial channel — is
>   untouched (and untouchable) by network rules. An escape is host code exec, game over for this layer.
>   - The host as a confused deputy. Host processes that act on guest-influenced data and then talk to the network are potential
>   low-bandwidth exfil channels. Concrete example already on the box: the CloudWatch collector ships pvestatd metrics — VM names, per-VM
>   resource counters — off-host. A guest can modulate its own CPU/disk usage and leak bits through metrics. Silly-sounding, real category;
>   bandwidth is tiny, but for eval-integrity threat models "tiny" isn't always "irrelevant."
> 
>   Operational failure modes (the control eats itself)
> 
>   - Fails open. Marker present + oneshot unit fails at boot (empty MGMT_NIC from a network-online race, say) → host runs unrestricted,
>   unit sits quietly in failed, samples run anyway. Nothing in the plugin asserts "lockdown is actually live" before starting a sample.
>   This is the gap I'd close first — it's cheap and it converts the control from "hopefully on" to "verified on."
>   - On-host tamper is silent. iptables -t mangle -F by any tool, script, or well-meaning admin reopens egress with no alarm. Same for
>   deleting the marker, which is just a root-writable file.
>   - The resolv-file blanking can be overwritten. Solid on EC2 (tmpfiles creates it once per boot, nothing else manages it). On the
>   virtualized build there's no tmpfiles entry — if resolvconf/DHCP hooks own that file, a lease renewal rewrites it and DNS egress
>   quietly comes back. The uid-owner OUTPUT drop I suggested in the review closes this properly on both variants.
>   - Rules are bound to one NIC name, captured at service start. A second interface with a route (an extra ENI attached later, a
>   VPN/tunnel interface someone brings up on the host) is an unblocked forwarding path. On EC2 as provisioned it's single-NIC, so this is
>   a drift risk rather than a current hole.
>   - Boot-time window. Between guest-start and unit-run there's theoretically an open moment — negligible here because the plugin creates
>   VMs via the API well after boot and nothing runs onboot=1, but it stops being negligible if that ever changes.
> 
>   What would close these at the host level
> 
>   Roughly in value order: (1) plugin-side verification — a config flag that makes sample_init check the rules and dnsmasq state on the
>   host and refuse to run otherwise, which converts fail-open, tamper, and unit-failure into loud failures at once; (2) the uid-owner
>   OUTPUT drop for dnsmasq; (3) fail-closed behaviour in the script (no NIC determined → blanket FORWARD drop, not exit 1).
> 
>   The middle group — dnsmasq exploits, VM escapes, confused-deputy channels — is the part no amount of host config can fully fix, because
>   the defence would have to survive the host being compromised. That's precisely the argument for the no-egress VPC/SG work: it's the
>   only layer in this stack whose enforcement point the guest (or an escaped guest) can't reach.

dpolatajko-aisi:
@jon-aisi I updated the scripts and added some tests based on your comments. I also manually tested a bunch of stuff against a live nitrovirt instance with Claude:

- Inert by default: no marker → service runs clean, zero rules, resolv.conf untouched — including after a reboot
- Apply: marker + start → both FORWARD drops at the top of the mangle chain on the management NIC, dnsmasq uid-owner OUTPUT drop, resolv.conf blanked with backup created
- Repair: manually deleted our OUTPUT rules and inserted a junk ACCEPT at position 1 → timer restored our rules above the junk within one tick, under a fresh RUN_ID
- No accumulation: tagged-rule count stayed exactly 3 across multiple timer ticks (GC works)
- Guest e2e (new pytest, env-gated): from inside a sandbox VM under lockdown — DHCP works, external IP unreachable, external DNS resolution fails, internal SDN DNS still answers
- Reboot with marker present: host came back locked down (rules + blanked resolv.conf) with no human intervention
- Fail-deadly: made the script fail → unit failure triggered the halt unit, pveproxy/pvedaemon stopped and runtime-masked, port 8006 confirmed dead from another machine, SSH unaffected
- Recovery: the README's documented unmask/start sequence brought the API back
- Unlock: marker removed → all rules gone, original resolver restored from backup, and the host passes the normal integration suite afterwards (no residue)
- Skip gate: without PROXMOX_EGRESS_LOCKDOWN_ENABLED the e2e test skips, so it can't break normal integration runs

artbristol:
I might be missing some relevant context, but I can't understand the difference between this change, vs disabling SNAT on the vnet (just a config change, no extra scripts needed?)

dpolatajko-aisi:
> I might be missing some relevant context, but I can't understand the difference between this change, vs disabling SNAT on the vnet (just a config change, no extra scripts needed?)

disabling SNAT still leaves DNS resolution open to the guest. also, this approach enables blocking egress on the fly as it doesn't require setting at the eval definition. also snat=false is still forwarding traffic off the host, and dropping it based on IP check in the ENI because guest traffic doesn't get its IP masked as the hosts, whereas this is dropping packets at the host forwarding step so it seems more strictly an "on-proxmox" egress denial. i think they complement each other though and defense in depth is the main idea here anyway

CONTRIBUTING.md
```diff
@@ -77,6 +77,27 @@ export PROXMOX_WINDOWS_TEMPLATE_TAG=<your-tag>
 
 With this set, tests in `test_proxmox_sandbox_agent_commands.py` will run for both Linux and Windows.
 
+### Egress lockdown test
+
+`test_egress_lockdown_e2e.py` is skipped unless `PROXMOX_EGRESS_LOCKDOWN_ENABLED`
```
danl-aisi, line 82: nit: does this test exist?

CONTRIBUTING.md
```diff
@@ -77,6 +77,27 @@ export PROXMOX_WINDOWS_TEMPLATE_TAG=<your-tag>
...
+
+The built-in VM template must already exist on the host before locking down:
+baking it boots the source image and installs packages (including
+`qemu-guest-agent`) from inside the guest, which needs guest egress. Any test
+that creates a default sandbox (e.g. `test_host_isolation_e2e.py`) bakes it on
+first use. Then:
+
+```bash
+ssh root@<proxmox-host> 'touch /etc/inspect-proxmox-egress-lockdown && systemctl start inspect-proxmox-egress-lockdown.service'
+PROXMOX_EGRESS_LOCKDOWN_ENABLED=1 uv run pytest tests/proxmoxsandboxtest/test_egress_lockdown_e2e.py
```
danl-aisi, line 94: nit: What does `PROXMOX_EGRESS_LOCKDOWN_ENABLED` do? I can't see it anywhere else

danl-aisi (approved):

CONTRIBUTING.md
danl-aisi: I think you mentioned tests in your response to @jon-aisi, did you push them?

dpolatajko-aisi: ah sorry you're right i forgot to commit the tests !

PR #109, opened 2026-08-10
Don't wait for each VM to boot before creating the next

bmillwood-aisi:
create_sdn_and_vms already awaits every VM once they have all been created, so awaiting inside clone_vm_and_start as well just serialised the boots. Now each VM's boot overlaps with the next VM's clone and config.

Creation itself stays sequential: /cluster/nextid is read-then-use and TaskWrapper waits for all new cluster tasks, so concurrent creates would need additional care.

This implementation deliberately chosen to be minimal and safe. I think VM destruction should also get this treatment, but I've chosen not to stick my nose in that at this time.

Seemed OK to make a breaking API change since we're already making some for the next release anyway. (Although actually now that I think of it, I haven't checked if any of the changed things are exported in a non-underscore-prefixed module.)

# Testing

- Existing tests all pass, including the `req_proxmox` tests, except I haven't managed to run the no-internet test yet.
- On a range I'm working on, this took startup from ~15 minutes to ~5 minutes, and it all seemed to start up okay.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 - Security: redact Proxmox passwords in configuration representations and validation error messages, and omit credentials from cleanup logs
 - Fix: `exec()` no longer aborts the sample when a command kills its own command-runner wrapper process
 - Fix: "500 QEMU guest agent is not running" is retried for much longer (~45s -> 8m25s)
-- Don't wait for a VM to reach "running" before starting the next one (just wait for all of them together at the end)
+- Don't wait for a VM to reach "running" before starting the next one (just wait for all of them together at the end). Set `await_before_next_vm=True` on a `VmConfig` if later VMs depend on it having booted first
 
 ## 0.11.0 - 2026-06-01
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Security: redact Proxmox passwords in configuration representations and validation error messages, and omit credentials from cleanup logs
 - Fix: `exec()` no longer aborts the sample when a command kills its own command-runner wrapper process
 - Fix: "500 QEMU guest agent is not running" is retried for much longer (~45s -> 8m25s)
+- Don't wait for a VM to reach "running" before starting the next one (just wait for all of them together at the end)
 
 ## 0.11.0 - 2026-06-01
 
--- a/tests/proxmoxsandboxtest/test_qemu_commands.py
+++ b/tests/proxmoxsandboxtest/test_qemu_commands.py
@@ -37,7 +37,7 @@ async def test_simple_vm_non_sandbox(
             uefi_boot=False,
         ),
         built_in_vm_ids=await built_in_vm.known_builtins(),
-        wait_until_ready=True,
+        wait_until_ready=False,
     )
 
     new_vm = await qemu_commands.read_vm(new_vm_id)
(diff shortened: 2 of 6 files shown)


bmillwood-aisi:
I think it's plausible that some ranges will previously have been relying on VMs starting up and being ready in a certain order, and that will now be broken by this. I'd still be excited to get something like this out, but maybe we should introduce an escape hatch.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Security: redact Proxmox passwords in configuration representations and validation error messages, and omit credentials from cleanup logs
 - Fix: `exec()` no longer aborts the sample when a command kills its own command-runner wrapper process
 - Fix: "500 QEMU guest agent is not running" is retried for much longer (~45s -> 8m25s)
+- Don't wait for a VM to reach "running" before starting the next one (just wait for all of them together at the end)
```
dpolatajko-aisi:
```suggestion
- Make waiting for a VM to reach "running" before starting the next one optional, and default this option to `False` (i.e. boot VMs concurrently and just wait for all VMs to be running at the end)
```

tests/proxmoxsandboxtest/test_qemu_commands.py
```diff
@@ -37,6 +37,7 @@ async def test_simple_vm_non_sandbox(
             uefi_boot=False,
         ),
         built_in_vm_ids=await built_in_vm.known_builtins(),
+        wait_until_ready=True,
```
dpolatajko-aisi: do the tests fail if this is `False`? if not, imo it would be better to test the default. if so, I think it'd be good to add awaits to the tests where relevant or add a new test covering the new behaviour

dpolatajko-aisi: (apply this ^ comment to all the tests in the file)

dpolatajko-aisi (approved):
looks good to me - i've a couple of nitpicks regarding testing and the changelog but won't be offended if you just want to get this in

CHANGELOG.md
bmillwood-aisi: I think in this edit you're imagining that people are using the underlying library directly -- they definitely could, which is e.g. why I kept `start_and_await` instead of removing it, but I expect they usually won't? so I was pitching my explanation at people who only interact with the inspect sandbox configuration

tests/proxmoxsandboxtest/test_qemu_commands.py
bmillwood-aisi: I switched most of them to `False`. Claude believes the ones left require `True` and has left a comment explaining why.

bmillwood-aisi:
sorry I decided to add a `VmConfig` field so that people could sequence their boots if they wanted to -- can you check you're happy with the new thing?

tests/proxmoxsandboxtest/test_qemu_commands.py
bmillwood-aisi: (I reran all the tests and they passed)

dpolatajko-aisi (approved):

CHANGELOG.md
dpolatajko-aisi: very fair - I like the CHANGELOG to straddle documenting what changed for a user and a developer, but it's just personal preference. your new edit seems good to me

PR #110, opened 2026-08-11
feat: enable extra headers in requests to Proxmox API

dpolatajko-aisi:
## Description

This PR enables the sandbox provider to pass extra headers in requests to the Proxmox PVE API. The motivating example is when the API is targeted via a reverse proxy which imposes its own auth header - this change allows the sandbox provider to pass such a header and successfully authenticate with the proxy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,28 +226,7 @@ Instances with the same `pool_id` form a pool. Each eval sample acquires one ins
 
 ### Extra HTTP Headers
 
-`extra_headers` on an instance adds HTTP headers to every request sent to that instance's Proxmox API, including file uploads. Typical uses: credentials for a proxy or access gateway in front of the API (a bearer token, Cloudflare Access service token, or similar), an API key for a gateway that routes or rate-limits on one, or tracing headers for request attribution:
-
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -224,6 +224,33 @@ export PROXMOX_CONFIG_FILE=/path/to/instances.json
 
 Instances with the same `pool_id` form a pool. Each eval sample acquires one instance from its pool, uses it exclusively, and releases it back when done. Concurrency is automatically limited to the total number of instances.
 
+### Extra HTTP Headers
+
+`extra_headers` on an instance adds HTTP headers to every request sent to that instance's Proxmox API, including file uploads. Typical uses: credentials for a proxy or access gateway in front of the API (a bearer token, Cloudflare Access service token, or similar), an API key for a gateway that routes or rate-limits on one, or tracing headers for request attribution:
```
art-dsit: I think we should just signpost schema.py here rather than bloat this already too-long base README.

art-dsit (approved):
LGTM, probably needs a short line in the changelog

PR #112, opened 2026-08-14
Move image_storage to instance config

bmillwood-aisi:
The way I see it, ProxmoxSandboxEnvironmentConfig is "what does the eval want the proxmox instance to configure so that it can run". The eval doesn't care what image storage the instance is using. Meanwhile, ProxmoxInstanceConfig is something your provisioning layer fills in with "these are the details of the instances I provisioned, that you'll need in order to use them properly", and image_storage is such a thing just like host and port are.

Benefits:
- evals no longer have to specify this thing they don't care about
- instance provisioning *can* set this and obviate the need to use PROXMOX_IMAGE_STORAGE in some cases
- fix a bug where if two evals ran on the same host / port / node, only the image_storage specified by the first would be used to create the infra_commands that both would use (I doubt this came up much).

I haven't thought too hard about migration / compatibility. The situation is a bit worse than it was in #103 because this field was actually useful at its old location. An alternative could be to keep this on `ProxmoxSandboxEnvironmentConfig` for now in some kind of deprecated way, and then somehow fall back to that if it's not specified on the instance, but that sounds somewhat fiddly so I'd be a bit inclined to simply eat the breakage. People who are typechecking their code should notice it that way.

## Testing

- [x] CI.
- [x] Manually running req_proxmox tests

```
======================= 203 passed, 5 skipped, 16 warnings in 1959.00s (0:32:39) =======================
```

### DIFF
--- a/tests/proxmoxsandboxtest/test_config_loading.py
+++ b/tests/proxmoxsandboxtest/test_config_loading.py
@@ -28,6 +28,7 @@ def test_load_instances_from_file():
                 "password": "secret",
                 "node": "pve1",
                 "verify_tls": False,
+                "image_storage": "fast-storage",
             },
             {
                 "instance_id": "test-2",
@@ -70,6 +71,7 @@ def test_load_instances_from_file():
         assert instances[0].password.get_secret_value() == "secret"
         assert instances[0].node == "pve1"
         assert instances[0].verify_tls is False
+        assert instances[0].image_storage == "fast-storage"
 
         assert instances[1].instance_id == "test-2"
         assert instances[1].pool_id == "kali-pool"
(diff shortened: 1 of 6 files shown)


tests/proxmoxsandboxtest/test_config_loading.py
```diff
@@ -248,24 +251,6 @@ def test_missing_required_fields_in_instance():
         os.environ.update(old_env)
 
 
-def test_sandbox_config_defaults():
```
bmillwood-aisi, line 251: The story here is that Codex added a line to this test that seemed not useful to me, so I went to look at it in more detail and decided the rest of the test also seemed not useful, and the test below it. Of course, there's nothing comprehensive about this, it's just the examples I happened to notice.

art-dsit (approved):
Make sense to me, thanks. I think we can risk a breaking change here

PR #116, opened 2026-08-20
Fix `sandbox_service()` on this provider: clamp reads to PVE's 16 MiB `agent/file-read` cap

art-dsit:
`sandbox_service()` has never worked on this provider, for any request size. It wraps
every request read in `override_max_exec_output_size(SERVICE_REQUEST_READ_OUTPUT_LIMIT)`
— 150 MiB — and PVE rejects an `agent/file-read` whose `count` exceeds 16 MiB, so every
poll 400s and no request is ever read. This clamps every read count to the PVE cap.

<details><summary>Mechanism and A/B verification</summary>

`SERVICE_REQUEST_READ_OUTPUT_LIMIT` (`inspect_ai/util/_sandbox/service.py:50`) is
`150 * 1024**2`. Its only caller is `_handle_request` at the same file's line 310, which
applies it unconditionally to each request read — the override is not conditional on
payload size, so a 40-byte request fails exactly like a 40 MiB one.

The failure is also silent-ish by construction. `_handle_request` only recovers from
`OutputLimitExceededError` (via `_discard_unreadable_request`, which at least unblocks the
client with an error). PVE's 400 isn't that, so the handler just raises, the request file
stays in the queue, and it is retried on every poll for as long as the sample lives.

Verified A/B on a bare `sandbox_service` — one `echo` method, guest calls it once through
the generated client, no bridge, no agent, no model calls, `inspect_ai` 0.3.239:

| | result |
|---|---|
| patched (this PR) | `GUEST_STDOUT=echo:ping`, method invoked, 1m25s |
| unpatched `main` | service starts, then 36 × `warning: Error handling sandbox service request: HTTP response error: 400 Parameter verification failed … value must have a maximum value of 16777216`; guest client never gets a response; sample dies `RuntimeError('Command timed out')` |

Same script, same host, only the provider file differs. So this is not specific to any one
consumer of `sandbox_service()` — `sandbox_agent_bridge` was just where I first hit it.

Unit-level, the regression test fails on 3 of 6 cases without the clamp
(`assert 157286400 <= 16777216`). The clamp is a ceiling only — a limit *below* 16 MiB
still wins, which `test_low_limit_is_still_honoured` pins down, since any read count
derived from an Inspect limit needs clamping without losing a lower caller-set limit.

Footgun for later: `read_file` has its own reason to respect the same cap, so the two
`16777216` literals there now share the constant. If PVE ever raises the limit, it's one
place.

</details>

### DIFF
--- a/src/proxmoxsandbox/_proxmox_sandbox_environment.py
+++ b/src/proxmoxsandbox/_proxmox_sandbox_environment.py
@@ -62,8 +62,8 @@ _IN_GUEST_KILL_GRACE = 5
 # PVE rejects an agent/file-read with `count` above this ("value must have a
 # maximum value of 16777216"), so every read count must be clamped to it —
 # including ones derived from Inspect limits, which callers can raise above it.
-# override_max_exec_output_size() does exactly that when the sandbox service
-# reads a bridge request, which used to 400 every poll and stall the bridge.
+# sandbox_service() raises MAX_EXEC_OUTPUT_SIZE to 150 MiB for every request
+# read (override_max_exec_output_size), which used to 400 on each poll.
 _PVE_AGENT_FILE_READ_MAX = 16 * 1024**2
 _PVE_AGENT_FILE_READ_MAX_STR = "16 MiB"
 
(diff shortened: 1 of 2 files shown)


art-dsit:
> I wasn't aware of the _sandbox_ agent bridge before reading this PR. It's unclear to me whether this is a surface we should be redteaming alongside proxmox this week, what do you think?

I'm including it in the scope of Inspect core red-teaming as I think any vulns would also apply to Docker and other sandboxes

> * It looks like you typically include changelog updates in PRs to this repo, do you need one here?

Fixed

PR #117, opened 2026-08-21
feat: support attaching to existing sandbox

wbarnes-aisi:
This is only a PoC - I suspect this code is not production ready (in particular I have concerns that the "existing sandbox overrides the spinning up sandbox" structure is fragile compared to having them both inherit from an abstract base class)  - but I want it to be committed onto a branch as it is regardless a useful intermediate step that I'll be reusing.

See CASTCE-5953 for context

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,7 @@ dependencies = [
 
 [project.entry-points."inspect_ai"]
 "proxmox-sandbox" = "proxmoxsandbox._proxmox_sandbox_environment"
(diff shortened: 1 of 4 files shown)


wbarnes-aisi:
lol, definitely needs further work for productionising given ruff is complaining about badly formatted Python!

rather than tidying this (which might be polishing a bad design; thanks to modern language models for the ability to one-shot code like this) I'll bring up for further discussion as and when I end up using this code again

vhong-aisi:
@wbarnes-aisi should we convert this to draft if it's not yet ready?

